PR #2, opened 2020-06-23
Change the way time ranges are stored in sections from array to ordered list.

maceonthompson:
This changes TimeRange to work over a week long period, as opposed to a 24h period. I also added some helpful functions re: being over a week-long period to TimeRange.
It'll be useful to add some more checks, both in the section and in the frontend, for if the user inputs an impossible section/class (like the check in the `Section.java` constructor for overlapping class times within the same section). This could include having an unreasonably long meeting time (something like a 12 hours long etc).

### DIFF
--- a/src/main/java/com/google/sps/data/Section.java
+++ b/src/main/java/com/google/sps/data/Section.java
@@ -14,20 +14,26 @@
 
 package com.google.sps.data;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
 public class Section {
     private String professor;
-    private TimeRange[] meetingTimes;
-    /*
-    * meeting times is an array
-    * Where the 0th index is Sunday and the 6th is Saturday
-    * [_,_,_,_,_,_,_] no meeting times
-    * [_,*,_,*,_,*,_] class MWF
-    * [_,_,*,_,*,_,_] class TTh
-    * [*,_,_,_,_,_,*] class SatSun
-    */
-    public Section(String professor, TimeRange[] meetingTimes){
+    private List<TimeRange> meetingTimes;
+    // Meeting times will be sorted in ascending order of time, 
+    // with the first lesson being first in the list
+
+    public Section(String professor, List<TimeRange> meetingTimes){
         this.professor = professor;
         this.meetingTimes = meetingTimes;
+        Collections.sort(this.meetingTimes, TimeRange.ORDER_BY_START);
+
+        for(int i = 0; i < meetingTimes.size() - 1; i++){
+          if(meetingTimes.get(i).overlaps(meetingTimes.get(i + 1))) {
+            throw new IllegalArgumentException("A section's lecture times can't overlap!");
+          }
+        }
     }
 
     /**
--- a/src/main/java/com/google/sps/data/Section.java
+++ b/src/main/java/com/google/sps/data/Section.java
@@ -19,9 +19,9 @@ import java.util.List;
 
 public class Section {
     private String professor;
-    private List<TimeRange> meetingTimes;
     // Meeting times will be sorted in ascending order of time, 
     // with the first lesson being first in the list
+    private List<TimeRange> meetingTimes;
 
     public Section(String professor, List<TimeRange> meetingTimes){
         this.professor = professor;
--- a/src/main/java/com/google/sps/data/Section.java
+++ b/src/main/java/com/google/sps/data/Section.java
@@ -14,7 +14,6 @@
 
 package com.google.sps.data;
 
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
--- a/src/main/java/com/google/sps/data/TimeRange.java
+++ b/src/main/java/com/google/sps/data/TimeRange.java
@@ -186,6 +186,12 @@ public final class TimeRange {
     return a.start == b.start && a.duration == b.duration;
   }
 
+  /**
+   * Returns the time in minutes given a day, for use when creating new timeRanges.
+   * @param hours an int representing the time (in hours) during the day
+   * @param minutes an int representing the time (in minutes) during the hour
+   * @return time in minutes during a week
+   */
   public static int getTimeInMinutes(int hours, int minutes) {
     if (hours < 0 || hours >= 24) {
       throw new IllegalArgumentException("Hours can only be 0 through 23 (inclusive).");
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -21,10 +21,40 @@
...
+  public static final int START_OF_THURSDAY = getTimeInMinutes(THURSDAY, 0, 0);
+  public static final int END_OF_THURSDAY = getTimeInMinutes(THURSDAY, 23, 59);
+
+  public static final int START_OF_FRIDAY = getTimeInMinutes(FRIDAY, 0, 0);
+  public static final int END_OF_FRIDAY = getTimeInMinutes(FRIDAY, 23, 59);
+
+  public static final int START_OF_SATURDAY = getTimeInMinutes(SATURDAY, 0, 0);
+  public static final int END_OF_SATURDAY = getTimeInMinutes(SATURDAY, 23, 59);
+
+  public static final int START_OF_WEEK = getTimeInMinutes(0, 0);
```
hjohns28, line 53: should this be getTimeInMinutes(0,0,0)?

hjohns28 (approved):

src/main/java/com/google/sps/data/Section.java
```diff
@@ -14,20 +14,25 @@
...
-    * Where the 0th index is Sunday and the 6th is Saturday
-    * [_,_,_,_,_,_,_] no meeting times
-    * [_,*,_,*,_,*,_] class MWF
-    * [_,_,*,_,*,_,_] class TTh
-    * [*,_,_,_,_,_,*] class SatSun
-    */
-    public Section(String professor, TimeRange[] meetingTimes){
+    private List<TimeRange> meetingTimes;
+    // Meeting times will be sorted in ascending order of time, 
+    // with the first lesson being first in the list
```
kristen-law, line 23: Comments should generally go above the code it's referring to.

src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -168,6 +198,67 @@ public static int getTimeInMinutes(int hours, int minutes) {
     return (hours * 60) + minutes;
   }
 
+  public static int getTimeInMinutes(int day, int hours, int minutes){
```
kristen-law: Can you add a comment explaining what this does? It's not really clear why you want to convert a day & time into minutes.

kristen-law (changes requested):

src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -168,6 +198,67 @@ public static int getTimeInMinutes(int hours, int minutes) {
...
+    if (hours < 0 || hours > 24 * 7) {
+      throw new IllegalArgumentException("Hours must be within a week period (between 0 and 167, inclusive).");
+    }
+
+    if (minutes < 0 || minutes >= 60) {
+      throw new IllegalArgumentException("Minutes can only be 0 through 59 (inclusive).");
+    }
+
+    return (int) Math.floorDiv(hours, 24);
+  }
```
FabriceCastel: You can replace most of this function with a call to `getDayOfWeekHours()` to avoid code duplication

src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -168,6 +198,67 @@ public static int getTimeInMinutes(int hours, int minutes) {
...
+
+  /**
+   * Returns the day of the week in int form (0 being sunday, 6 being saturday)
+   * @param minutes
+   * @return int representing the day of the week
+   */
+  public static int getDayOfWeekMinutes(int minutes) {
```
FabriceCastel: This group of converter functions might be better named eg. `minutesToDayOfWeek`, `convertMinutesToWeekday` or something in that vein, to convey the fact that these functions take in some minute/hour/etc amount and convert it to a different unit. Typically `getXYZ` type names are reserved for getter methods that return some property of the object you're invoking it on.

src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -21,10 +21,40 @@
...
+  public static final int END_OF_THURSDAY = getTimeInMinutes(THURSDAY, 23, 59);
+
+  public static final int START_OF_FRIDAY = getTimeInMinutes(FRIDAY, 0, 0);
+  public static final int END_OF_FRIDAY = getTimeInMinutes(FRIDAY, 23, 59);
+
+  public static final int START_OF_SATURDAY = getTimeInMinutes(SATURDAY, 0, 0);
+  public static final int END_OF_SATURDAY = getTimeInMinutes(SATURDAY, 23, 59);
+
+  public static final int START_OF_WEEK = getTimeInMinutes(0, 0);
+  public static final int END_OF_WEEK = getTimeInMinutes(6, 23, 59);
```
FabriceCastel: I wager you could greatly reduce the need for these `START_OF.../END_OF...` constants by updating the `fromStartDuration` method on this class to better match your use case.

Something like..

```
public static TimeRange fromStartDuration(int day, int hour, int minute, int durationMinutes) { ... }
```

Same for `fromStartEnd`. It'll likely make your life easier.

FabriceCastel (changes requested):

src/main/java/com/google/sps/data/TimeRange.java
maceonthompson: Added comment (and some other docs clarifying why the function is important). Both constructors for the TimeRange Object take arguments in minutes, so converting to minutes given a day & time makes that easier.
Though as I'm writing this I'm realizing it might make more sense to modify the constructors - I'll talk about it/ask about it during the daily sync.

src/main/java/com/google/sps/data/Section.java
maceonthompson: Resolved!

src/main/java/com/google/sps/data/TimeRange.java
maceonthompson: Resovled!

src/main/java/com/google/sps/data/TimeRange.java
maceonthompson: Resolved!

maceonthompson:
I'm not sure what the build check failure means here, it could have something to do with the fact that I switched over to working locally on the chromebook? The error doesn't mention an actual build failure, so it should be compiling alright - it says that there isn't an associated cloud project, which there definitely should be.

Edit: the error message is 
> This repository is not mapped to a Google Cloud Platform project. Please connect it to a project in the cloud console.

FabriceCastel:
> I'm not sure what the build check failure means here, it could have something to do with the fact that I switched over to working locally on the chromebook? The error doesn't mention an actual build failure, so it should be compiling alright - it says that there isn't an associated cloud project, which there definitely should be.
> 
> Edit: the error message is
> 
> > This repository is not mapped to a Google Cloud Platform project. Please connect it to a project in the cloud console.

For the time being you can pretend that doesn't exist. Kristen has been looking into this, so we'll either disable it or have it do something useful rather than be a broken bot.

FabriceCastel (approved):
LGTM

kristen-law (approved):

PR #4, opened 2020-06-23
Switch from invididual invariant arguments to an "Invariants" class

maceonthompson:
Re: advice from Fabrice

### DIFF
--- a/src/main/java/com/google/sps/data/Invariants.java
+++ b/src/main/java/com/google/sps/data/Invariants.java
@@ -1,0 +1,19 @@
+package com.google.sps.data;
+
+public class Invariants {
+  private float minCredits;
+  private float maxCredits;
+
+  public Invariants(float minCredits, float maxCredits){
+    this.minCredits = minCredits;
+    this.maxCredits = maxCredits;
+  }
+
+  public float getMinCredits(){
+    return this.minCredits;
+  }
+
+  public float getMaxCredits(){
+    return this.maxCredits;
+  }
+}
--- a/src/main/java/com/google/sps/data/Invariants.java
+++ b/src/main/java/com/google/sps/data/Invariants.java
@@ -4,16 +4,16 @@ public class Invariants {
   private float minCredits;
   private float maxCredits;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/data/Invariants.java
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  public float getMinCredits(){
+    return this.minCredits;
+  }
+
+  public float getMaxCredits(){
+    return this.maxCredits;
+  }
+}
```
achillesecos: New line here.

maceonthompson: Resolved!

achillesecos (commented):
Just wondering, are you going to include the other 3 invariants here? 

(From Design Doc)
- 1.  No courses will overlap in a schedule
- 2.  The total number of credits contained in a schedule must be within the min/max bound of credits (specified by the user)
- 3.  If a course is marked as required, it must be in the schedule.
- 4. There can’t be multiple sections of the same course in the same schedule

Seems like you check invariant 4 in Section.java, will you check for the other invariants here?

maceonthompson:
> Just wondering, are you going to include the other 3 invariants here?
> 
> (From Design Doc)
> 
> * 1. No courses will overlap in a schedule
> * 1. The total number of credits contained in a schedule must be within the min/max bound of credits (specified by the user)
> * 1. If a course is marked as required, it must be in the schedule.
> * 1. There can’t be multiple sections of the same course in the same schedule
> 
> Seems like you check invariant 4 in Section.java, will you check for the other invariants here?

No. This serves as a capsule for invariants that are changeable by the user (which for right now is the min/max number of credits). So it's more of a wrapper for actual data as opposed to different logical implementations. The ideas of no course overlap, same section overlap, and required courses are more changes in the logic of the `generateSchedules()` function in the scheduler class.

achillesecos:
Ok sounds good. I understand what you mean now.

achillesecos (approved):

kristen-law (approved):
LGTM

PR #7, opened 2020-06-25
Add basic testing for scheduler class

maceonthompson:
More testing (much, much more testing) will be added in the future. This is more of an initial pull request so there isn't a massive one down the line.


maceonthompson:
I'm gonna delete this and create a new branch, as the timeline will get pretty ugly if I try to merge in the current state.

PR #8, opened 2020-06-25
Add basic testing for Scheduler Class

maceonthompson:
More testing (much, much more testing) will be added in the future. This is more of an initial pull request so there isn't a massive one down the line.

### DIFF
--- a/src/test/java/com/google/sps/SchedulerTest.java
+++ b/src/test/java/com/google/sps/SchedulerTest.java
@@ -1,0 +1,98 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


hjohns28 (approved):
LGTM

src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -0,0 +1,98 @@
...
+        "Wrongdoing", 1, true, Arrays.asList(section));
+    
+    Collection<Schedule> actual = scheduler.generateSchedules(Arrays.asList(course1), 
+        new Invariants(3, 5));
+    Assert.assertEquals(Collections.emptyList(), actual);
+  }
+
+  @Test
+  public void twoCoursesValidSchedule() {
+    //TODO: This
```
FabriceCastel, line 89: A common pattern when working on group projects is to note who's on the hook for a specific TODO item, eg.

```
// TODO(fcastel) fix these shenanigans
```

This helps ensure you eventually do go back to these items (at least with a higher likelihood) rather than having everyone ignore the TODO because they don't feel personally liable for it.

FYI @achillesecos & @hjohns28

FabriceCastel (approved):
LGTM

PR #9, opened 2020-06-25
Implement Basic Schedule Permutation

maceonthompson:
Tests will currently fail because Schedules, Courses, and Sections don't have proper .equals() functions, and the final returned objects aren't strictly equal (different values in memory). But when debugging, the first basic test works!!

### DIFF
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -28,7 +28,7 @@ public final class Scheduler {
 
     generateSchedulesHelper(betterCourses, new ArrayList<>(), invariants, schedulesSet);
 
-    System.out.println(schedulesSet);
+    System.out.println(courses);
     List<Schedule> schedules = new ArrayList<>();
     for (List<Course> schedule : schedulesSet) {
       schedules.add(new Schedule(schedule));
@@ -105,9 +105,9 @@ public final class Scheduler {
   private boolean sectionsOverlap(Section toCheck, List<Section> sections) {
     for (Section section : sections) {
       if (toCheck.overlaps(section)) {
-        return false;
+        return true;
       }
     }
-    return true;
+    return false;
   }
 }
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -22,12 +22,13 @@ public final class Scheduler {
    * @return A list of valid schedules.
    */
   public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants) {
-
-    Collections.sort(courses, Course.REQUIRED_TO_FRONT);
+    ArrayList<Course> betterCourses = new ArrayList<>(courses);
+    Collections.sort(betterCourses, Course.REQUIRED_TO_FRONT);
     Set<List<Course>> schedulesSet = new HashSet<List<Course>>();
 
-    generateSchedulesHelper(courses, Arrays.asList(), invariants, schedulesSet);
+    generateSchedulesHelper(betterCourses, new ArrayList<>(), invariants, schedulesSet);
 
+    System.out.println(schedulesSet);
     List<Schedule> schedules = new ArrayList<>();
     for (List<Course> schedule : schedulesSet) {
       schedules.add(new Schedule(schedule));
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -1,6 +1,6 @@
 package com.google.sps;
 
-import java.util.Collection;
+import java.util.List;
 
 import com.google.sps.data.*;
 
@@ -8,7 +8,7 @@ import com.google.sps.data.*;
  * Static class for permuting schedules given a list of courses.
  */
 public final class Scheduler {
-    public Collection<Schedule> generateSchedules(Collection<Course> courses, Invariants invariants){
+    public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants){
         //TODO: This
         return null;
     }
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -1,15 +1,112 @@
 package com.google.sps;
 
-import java.util.List;
-
 import com.google.sps.data.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
- * Static class for permuting schedules given a list of courses.
+ * Static class for permuting schedules given a list of courses and a set of Invariants.
  */
 public final class Scheduler {
+  /**
+   * Generates all valid schedules given a list of courses and invariant
+   * constraints.
+   * 
+   * @param courses    The list of courses to generate the schedules from.
+   * @param invariants The invariants that constrain the valid schedules
+   * @return A list of valid schedules.
+   */
   public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants) {
-    // TODO: This
-    return null;
+
+    Collections.sort(courses, Course.REQUIRED_TO_FRONT);
+    Set<List<Course>> schedulesSet = new HashSet<List<Course>>();
+
+    generateSchedulesHelper(courses, Arrays.asList(), invariants, schedulesSet);
+
+    List<Schedule> schedules = new ArrayList<>();
+    for (List<Course> schedule : schedulesSet) {
+      schedules.add(new Schedule(schedule));
+    }
+
+    return schedules;
+  }
+
+  /**
+   * Recursively iterates through a list of courses and adds schedules to the
+   * given set.
+   * 
+   * @param courses    The list of courses to iterate through
+   * @param schedule   The current schedule to be added/operated on
+   * @param invariants The invariants that constrain the valid schedules
+   * @param schedules  The set of schedules that is being generated/added to.
+   */
+  private void generateSchedulesHelper(List<Course> courses, List<Course> schedule, 
+      Invariants invariants, Set<List<Course>> schedules) {
+
+    // This means that we have a valid schedule, and should add it to the list.
+    float currCredits = totalCredits(schedule);
+    if (currCredits >= invariants.getMinCredits() && currCredits <= invariants.getMaxCredits()) {
+      schedules.add(schedule);
+    }
+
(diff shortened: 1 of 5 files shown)


maceonthompson:
This is a pretty chunky pull request, so I'm sorry about that! It also doesn't take into account overlapping required courses/the invariant that all required courses must be included in the schedule. But the steps to implement that shouldn't be *too* bad.
Edit: Also, if you're reviewing it and have questions re: what does what, feel free to ping me! I'd be happy to explain anything.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,33 +15,58 @@
...
-     */
-    public boolean overlaps(Section other){
-        //TODO: This
+    TimeRange thisCurr = thisTimes.removeFirst();
+    TimeRange otherCurr = otherTimes.removeFirst();
+
+    while (!thisTimes.isEmpty() && !otherTimes.isEmpty()) {
+      if (thisCurr.overlaps(otherCurr)) {
         return false;
+      } else if (thisCurr.end() > otherCurr.end()) {
```
hjohns28: I might be misunderstanding, but if thisCurr ends later than otherCurr, and the sections are sorted in ascending order, shouldn't you move on to the next (later) section in otherTimes, not thisTimes? (ie if thisCurr is 10-11, and otherCurr is 8-9 (so 11 > 9), if you move onto a later thisTimes section, they will never overlap?) Sorry if I didn't follow the code right!

hjohns28 (changes requested):

src/main/java/com/google/sps/data/Section.java
maceonthompson: You're 100% correct, good catch.

hjohns28 (approved):
LGTM

maceonthompson:
Fixed .equals methods! So testing should now properly function. From here on out I'll have separate pull requests into this branch to avoid mega-chunk pull requests (Though when this eventually gets pulled into master I'm sure it'll be a bit of a headache).

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+  /**
+   * Generates all valid schedules given a list of courses and invariant
+   * constraints.
+   * 
+   * @param courses    The list of courses to generate the schedules from.
+   * @param invariants The invariants that constrain the valid schedules
+   * @return A list of valid schedules.
+   */
+  public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants) {
+    ArrayList<Course> betterCourses = new ArrayList<>(courses);
```
FabriceCastel, line 25: `betterCourses` -> `sortedCourses` or `orderedCourses`

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+    Collections.sort(betterCourses, Course.REQUIRED_TO_FRONT);
+    Set<List<Course>> schedulesSet = new HashSet<List<Course>>();
+
+    generateSchedulesHelper(betterCourses, new ArrayList<>(), invariants, schedulesSet);
+
+    System.out.println(courses);
```
FabriceCastel, line 31: nit: remove the debug print statement

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+   * constraints.
+   * 
+   * @param courses    The list of courses to generate the schedules from.
+   * @param invariants The invariants that constrain the valid schedules
+   * @return A list of valid schedules.
+   */
+  public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants) {
+    ArrayList<Course> betterCourses = new ArrayList<>(courses);
+    Collections.sort(betterCourses, Course.REQUIRED_TO_FRONT);
+    Set<List<Course>> schedulesSet = new HashSet<List<Course>>();
```
FabriceCastel, line 27: extreme nit: `schedulesSet` -> `scheduleSet`

I don't think most people care about something like this if I'm honest, so feel free to ignore if you prefer it as is. When I've got a collection of foobars, I'll either use `foobars` or `foobarList`, `foobarSet`, etc. for variable names.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+
+    generateSchedulesHelper(betterCourses, new ArrayList<>(), invariants, schedulesSet);
+
+    System.out.println(courses);
+    List<Schedule> schedules = new ArrayList<>();
+    for (List<Course> schedule : schedulesSet) {
```
FabriceCastel, line 33: nit: `schedule` -> `courses` or `courseList`

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+   * @param schedule   The current schedule to be added/operated on
+   * @param invariants The invariants that constrain the valid schedules
+   * @param schedules  The set of schedules that is being generated/added to.
+   */
+  private void generateSchedulesHelper(List<Course> courses, List<Course> schedule, 
+      Invariants invariants, Set<List<Course>> schedules) {
+
+    // This means that we have a valid schedule, and should add it to the list.
+    float currCredits = totalCredits(schedule);
+    if (currCredits >= invariants.getMinCredits() && currCredits <= invariants.getMaxCredits()) {
```
FabriceCastel, line 54: nit-adjacent: you could improve legibility here by implementing a `meetsCreditRequirements(float credits)` method on the `Invariants` class.

```
if (invariants.meetsCreditRequirements(credits)) {
  ...
}
```

Is pretty well self-documenting. It would also be easy to extend this method to be more of a "does this schedule meet all of the requirements" function and keep the logic checking for this well contained rather than have it grow into an increasingly dense `if` statement.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+
+    /*
+     * Iterate over each section in the current course and check for overlap. If the
+     * section doesn't overlap and adding it would result in a valid schedule, add
+     * it and make a new recursive call.
+     */
+    // TODO: Implement logic for required courses having no valid schedules
+    for (Section section : courseSections) {
+      if (!sectionsOverlap(section, scheduleSections)
+          && currCredits + course.getCredits() <= invariants.getMaxCredits()) {
```
FabriceCastel, line 77: This `if` statement is a bit hard to read. It might be helpful to assign this boolean expression to a variable named after the condition it's checking for, eg.

```
bool doesNotOverlapSchedule = !sectionsOverlap(....);
bool doesNotExceedCreditLimit = currCredits + ........;

if (doesNotOverlapSchedule && doesNotExceedCreditLimit) {
  ...
}
```

is easier to understand at a glance

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,113 @@
...
+  /**
+   * Recursively iterates through a list of courses and adds schedules to the
+   * given set.
+   * 
+   * @param courses    The list of courses to iterate through
+   * @param schedule   The current schedule to be added/operated on
+   * @param invariants The invariants that constrain the valid schedules
+   * @param schedules  The set of schedules that is being generated/added to.
+   */
+  private void generateSchedulesHelper(List<Course> courses, List<Course> schedule,
```
FabriceCastel, line 49: Is there a reason for `schedule` not being a `Schedule` rather than a `List<Course>`? I'm asking this because your `totalCredits` method below looks like the sort of thing that would be defined in `Schedule` as a convenience method of sorts.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,33 +15,76 @@
...
+   * 
+   * @param other The other section.
+   * @return If there's overlap between the current section and the other section
+   */
+  public boolean overlaps(Section other) {
+    LinkedList<TimeRange> thisTimes = new LinkedList<>(this.meetingTimes);
+    LinkedList<TimeRange> otherTimes = new LinkedList<>(other.getMeetingTimes());
+
+    TimeRange thisCurr = thisTimes.removeFirst();
+    TimeRange otherCurr = otherTimes.removeFirst();
```
FabriceCastel, line 58: What happens here if a section has no meeting times?

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,33 +15,76 @@
...
-    /**
-     * Function to return whether or not a section overlaps with another
-     * @param other
-     * @return If there's overlap between the current section and the other section
-     */
-    public boolean overlaps(Section other){
-        //TODO: This
+    while (!thisTimes.isEmpty() && !otherTimes.isEmpty()) {
+      if (thisCurr.overlaps(otherCurr)) {
         return false;
```
FabriceCastel, line 62: Should this be `true` instead?

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,33 +15,76 @@
...
+    TimeRange otherCurr = otherTimes.removeFirst();
 
-    /**
-     * Function to return whether or not a section overlaps with another
-     * @param other
-     * @return If there's overlap between the current section and the other section
-     */
-    public boolean overlaps(Section other){
-        //TODO: This
+    while (!thisTimes.isEmpty() && !otherTimes.isEmpty()) {
```
FabriceCastel, line 60: This will not end up checking for all possible collisions. If you have a `thisTimes` with times, `A, B` and a `otherTimes` with time `C` like so:

```
A |--|
B       |----|
C        |----|
```

You'll take the first element of both lists, `A` and `C`, get a `false` when checking the condition for this while loop, then go on to the last `if` check between `A` and `C`, which will pass since they don't overlap. `B` and `C` will not have been tested here.

The edge case here is that if you end up hitting the end of one of the lists of `TimeRange`s, you stop checking for overlaps between the last element of the list you reached the end of and the other remaining items in the other list.

I would encourage you to write a unit test for this case to validate that it does indeed break, then to write the code that fixes that test.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,33 +15,76 @@
...
+  @Override
+  public boolean equals(Object other) {
+    if (other == this) {
+      return true;
+    } else {
+      return other instanceof Section && equals(this, (Section) other);
+    }
+  }
+
+  private boolean equals(Section a, Section b) {
```
FabriceCastel, line 86: This seems like it should be a static method since it doesn't rely on `this` at all

FabriceCastel (changes requested):

PR #10, opened 2020-06-29
Add hashCode() method to course, schedule, and section classes

maceonthompson:


hjohns28 (approved):
LGTM

src/main/java/com/google/sps/data/Section.java
```diff
@@ -35,6 +36,13 @@ public Section(String professor, List<TimeRange> meetingTimes){
         }
     }
 
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(professor, meetingTimes);
+  }
```
FabriceCastel: nit: please use consistent indentation

this can mean either indent this code some more to match the class, or indent the other functions in this class with two spaces rather than four

bonus option: indent your function two spaces more now then run a linter on everything to fix indentation everywhere and submit that as a separate style fixing pull request

FabriceCastel (approved):
LGTM with nit

PR #11, opened 2020-06-29
initial authorization to access / edit users' calendars

hjohns28:
(will later try to reduce scope of access from CALENDAR to CALENDAR_EVENTS, but isn't working at the moment)


maceonthompson (approved):
LGTM!

src/main/webapp/script.js
```diff
@@ -15,14 +15,6 @@
...
-  const greetings =
-      ['Hello world!', '¡Hola Mundo!', '你好，世界！', 'Bonjour le monde!'];
-
-  // Pick a random greeting.
-  const greeting = greetings[Math.floor(Math.random() * greetings.length)];
-
-  // Add it to the page.
-  const greetingContainer = document.getElementById('greeting-container');
-  greetingContainer.innerText = greeting;
+function userAuthorization() {
```
kristen-law, line 18: Small nit: This function name would be more clear if it's phrased as an action that describes what it's doing. I'd recommend renaming this to something like `authorizeUser`.

kristen-law (approved):
LGTM, just a small nit to fix before submitting.

PR #13, opened 2020-07-01
Add overlaps() functionality to Section class

maceonthompson:


src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,56 +15,112 @@
...
+        }
+      }
     }
 
+    // Accounts for the single element list case.
+    if (thisCurr.overlaps(otherCurr)) {
+      return true;
+    }
+
+    if (thisCurr.end() > TimeRange.END_OF_WEEK && saturdayOverlapsSunday(thisCurr, otherFirst)) {
```
maceonthompson: These final 3 if statements could be combined & written into the final `return` statement, something like 
```
return (this.curr.overlaps(otherCurr)) 
    || (thisCurr.end > TimeRange.END_OF_WEEK && SaturdayOverlapsSunday(thisCurr, otherFirst)) 
    || (otherCurr.end > TimeRange.END_OF_WEEK && SaturdayOverlapsSunday(otherCurr, thisFirst));
```
 but I thought it would be more legible to keep it as separate if statements. Let me know if it's best practice to combine them, however!

maceonthompson: The more I look at this the more I'm actually preferring the consolidated version... Let me know your thoughts but I might change it to this.

hjohns28 (approved):
LGTM

src/main/java/com/google/sps/data/Section.java
achillesecos: I think you could just ` return thisCurr.end() > TimeRange.END_OF_WEEK && saturdayOverlapsSunday(thisCurr, otherFirst)` since you check if its true, and return true. 

EDIT: actually I didnt see the below statement, nevermind.

src/test/java/com/google/sps/SectionTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class SectionTest {
+
+  private static final TimeRange MONDAY_10AM = 
+      TimeRange.fromStartDuration(TimeRange.MONDAY, 10, 0, 60);
```
achillesecos, line 18: Maybe might be worth writing a variable like `private static final int DURATION_1_HOUR = 60;` to avoid writing 60

src/main/java/com/google/sps/data/Section.java
hjohns28: I think the consolidated version is still legible, if you want to change it

src/test/java/com/google/sps/SectionTest.java
```diff
@@ -0,0 +1,110 @@
...
+    Assert.assertTrue(s1.overlaps(s2));
+  }
+
+  @Test
+  public void multipleElementsMidOverlaps() {
+    List<TimeRange> s1Times = Arrays.asList(MONDAY_10AM, MONDAY_1PM, TUESDAY_10AM);
+    List<TimeRange> s2Times = Arrays.asList(
+        TimeRange.fromStartDuration(TimeRange.MONDAY, 11, 30, 30),
+        TimeRange.fromStartDuration(TimeRange.MONDAY, 13, 30, 60),
+        TimeRange.fromStartDuration(TimeRange.SATURDAY, 15, 30, 60));
```
achillesecos: Also could be potentially good to define these above as `MONDAY_1130AM` and so on

maceonthompson: For now I'm leaving these as is, as they're only used once, and the code is pretty legible.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,56 +15,101 @@
...
+   * 
+   * @param other the section to check for overlap with.
+   * @return If there's overlap between the current section and the other section
+   */
+  public boolean overlaps(Section other) {
+    LinkedList<TimeRange> thisTimes = new LinkedList<>(this.meetingTimes);
+    LinkedList<TimeRange> otherTimes = new LinkedList<>(other.getMeetingTimes());
+
+    TimeRange thisCurr = thisTimes.removeFirst();
+    TimeRange otherCurr = otherTimes.removeFirst();
```
FabriceCastel: You should check for the case where one or both `TimeRange` lists are empty before fetching their respective first elements

src/main/java/com/google/sps/data/Section.java
```diff
@@ -15,56 +15,101 @@
...
+        } else {
+          otherCurr = otherTimes.removeFirst();
+        }
+      } else {
+        if (!otherTimes.isEmpty()) {
+          otherCurr = otherTimes.removeFirst();
+        } else {
+          thisCurr = thisTimes.removeFirst();
+        }
+      }
```
FabriceCastel: I was thinking about whether or not there was a more elegant way to write this out, and I think there might be. It relies on the assumption that a given section cannot have two `TimeRange` elements that overlap in its `meetingTimes`.

1. Create a combined `List<TimeRange>` that contains the `TimeRange` elements from both sections' `meetingTimes`
2. Sort that list by `TimeRange` starting time
3. Iterate through the list and check to see if any two successive entries overlap (make sure to check for the first & last elements here as well)

It might look something like this instead, if you wanted to try something a little different using the `Stream` API... (note, that's more for the sake of demonstration/illustration, you and the pod can decide if you prefer this type of style versus an imperative loop - my bias is clear, but that's all is it; bias :)

```
ArrayList<TimeRange> combinedMeetingTimes = (....);
Collections.sort(combinedMeetingTimes, ....);
int meetingCount = combinedMeetingTimes.size();

return IntStream.range(0, meetingCount + 1)
  .filter(index -> {
    TimeRange first = combinedMeetingTimes.get(index);
    TimeRange second = combinedMeetingTimes.get((index + 1) % meetingCount);
    return first.overlaps(second);
  })
  .findAny()
  .isPresent();
```

That's pretty close to the imperative version, too. It's a bit code golf-ey, could be made more legible with intermediate results getting assigned to name variables, but that's the idea in a nutshell.

For what it's worth, this approach handles the empty `TimeRange` list case for free.

FabriceCastel (changes requested):

src/main/java/com/google/sps/data/Section.java
FabriceCastel: Either approach is reasonable. I think it could be made more legible by naming the chunks of your expression. Another of my comments on this PR more or less invalidates the need for this compound expression though.

src/main/java/com/google/sps/data/Section.java
maceonthompson: This approach is super elegant, but  I'm going to go with the imperative version. The issue is that I still need to compare the first and last elements, but only want to actually check those if the combined list is of size > 1, which requires manually handling the empty case before other logic is applied. (This is specific to the Saturday ->  Sunday edge case)

maceonthompson: I'm definitely going to switch the code over to something more similar to this suggestion though, because event though it's technically "slower" (O(nLogn) due to sort vs O(2n) in the original implementation), it's far more legible.

src/main/java/com/google/sps/data/Section.java
maceonthompson: Added!

src/main/java/com/google/sps/data/Section.java
FabriceCastel: > The issue is that I still need to compare the first and last elements, but only want to actually check those if the combined list is of size > 1, which requires manually handling the empty case before other logic is applied.

The code above performs this check. The imperative version would check for that as well, assuming it's structured the same. This is true because the code runs N iterations, where N = size of the list. When it gets to the last iteration, you'll have `first` be the last element in the list and `second` be the first element in the list.

There is no need for additional checks/handling either; if the list is empty, no iterations will run and you won't find any overlap.

FabriceCastel: As for the running time, you're right that this is slower. The merging of the two already sorted lists isn't particularly smart in my example. If you really wanted to, you could write a O(n) merge algorithm and retain the O(n) overall runtime but I'm not convinced that's necessary given the use case here. Seems a tad overkill.

FabriceCastel: I've gone back and had a look at your Sunday > Monday edge case handling. It seems less like an edge case you should be dealing with here and more of an edge case `TimeRange` should handle for you out of the box when you call its `overlaps()` method.

```
TimeRange a = Sunday at 11pm to Monday at 6am
TimeRange b = Monday at 4am to Monday at 8am

a.overlaps(b) => true
```

This is how it should be, no? It seems like the current implementation would return false here instead when in fact, these two `TimeRange` objects do represent ranges of time that overlap.

This comes down to the `TimeRange::contains()` implementation, which will return false if you ask it whether the `TimeRange a` from above contains the moment in time on Monday at 1am. I think addressing the problem there will save you some pain and prevent that edge case from escaping and polluting other parts of your codebase which should not be concerned with the internals of a `TimeRange`.

maceonthompson: What happens in the case of a single element list? I.E. one of the lists are empty, and the other has a single element.

FabriceCastel: Fair point. This would pair up the element with itself since `index == (index + 1) % meetingCount` when `index = 0` and `meetingCount = 1`, which means that in this case it will find an overlap that isn't really there. Guess this can't escape having to check for the edge case where one or both lists are empty after all, not without some overly complicated or hacky shenanigans anyways. The easiest fix for the example code above then would be to throw a

```
if (meetingCount < 2) {
  return false;
}
```

in there before the stream.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -14,57 +14,94 @@
...
+   * overlaps with a TimeRange earlier in the week.
+   * 
+   * @param satEvent the event that starts on Sat but ends on Sunday
+   * @param earliestEvent the event earlier in the week.
+   * 
+   * @return a boolean representing if the two events overlap.
+   */
+  private boolean saturdayOverlapsSunday(TimeRange satEvent, TimeRange earliestEvent) {
+    return earliestEvent.contains(satEvent.end() - TimeRange.END_OF_WEEK);
+  }
```
FabriceCastel: Re-iterating this here so that it doesn't get lost in the other comment chain; this behavior belongs in `TimeRange`, not in `Section`.

maceonthompson: I didn't see those other comments! Thanks.

maceonthompson: Fixed!

src/main/java/com/google/sps/data/Section.java
maceonthompson: Addressed these! Ended up going with imperative because it felt a bit more idiomatic/legible. I still like the functional method though.

FabriceCastel (approved):
LGTM

PR #14, opened 2020-07-01
Rank schedule

achillesecos:
Parent: https://github.com/googleinterns/step78-2020/commit/eab24446cab792bce5fae9ad8b7a94a6af3e1f34

TLDR: Added the criteria preference comparator, and the 3 main criteria, with each criteria implementing the preference score interface (basic algorithm for now). Will add tests soon.

EDIT: Added tests. Will soon add multiple sections of same course tests.


achillesecos:
@googlebot I signed it!

achillesecos:
@googlebot I signed it!

src/main/java/com/google/sps/data/restrictTimesCriteria.java
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+    return minutesPenalty;
+  }
+
+  // Calculates the overlapping minutes between two TimeRanges
+  int calculateDifference(TimeRange a, TimeRange b) {
+    int start = Math.max(a.start(), b.start());
+    int end = Math.min(a.end(), b.end());
+    return end - start;
```
maceonthompson: This will always return a negative value (which i see that you took into account on line 32). It might be easier to have start be the Math.min and end be the Math.max, which is also in line with how TimeRanges deal with "earlier" and "later" times. (An earlier time is a lower number than a later time).

src/main/java/com/google/sps/data/prioritizeCoursesCriteria.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Arrays;
+import java.util.ArrayList;
+
+public class prioritizeCoursesCriteria implements Preference {
+
+  public void prioritizeCoursesCriteria(List<Course> courseList) {
+    this.courseList = courseList;
+  }
+
+  List<Course> courseList;
```
maceonthompson: nit: you can choose if you want to change this (according to the google style guide the way you're doing it is 100% valid), but Heidi and I both declare our fields before the constructor. It might be worth switching for style consistency throughout the project.

src/main/java/com/google/sps/data/Section.java
```diff
@@ -67,4 +67,9 @@ public boolean overlaps(Section other){
         //TODO: This
         return false;
     }
-}
\ No newline at end of file
+
+    // Getter for meetingTimes
+    public List<TimeRange> getMeetingTimes() {
```
maceonthompson: We might run into a merge conflict here, as I added the same function! But depending on which pull request gets merged first, I'll edit mine so that it shouldn't be too gnarly.

src/main/java/com/google/sps/Rank.java
```diff
@@ -1,11 +1,26 @@
...
-      
     return null;
   }
+
+  /**
+   * A comparator for sorting ranges by their start time in ascending order.
+   */
+  public static final Comparator<Schedule> ORDER_BY_CRITERIA = new Comparator<TimeRange>() {
+    @Override
+    public int compare(TimeRange a, TimeRange b) {
```
maceonthompson: This is declared as a comparator for Schedules but compares TimeRanges. Did you mean to update it to use a different compare method?

maceonthompson (changes requested):

src/main/java/com/google/sps/data/Section.java
achillesecos: Ah I see. Thats fine. I'll keep that in mind when resolving merge conflicts.

src/main/java/com/google/sps/data/restrictTimesCriteria.java
achillesecos: I'm essentially trying to find the difference that `a` and `b` overlap by, such that I update the minutesPenalty with that difference. I keep it negative because the more overlap means the less desirable it is.

src/main/java/com/google/sps/data/prioritizeCoursesCriteria.java
achillesecos: Ah ok thats fair. I'll do that to all the criteria classes.

src/main/java/com/google/sps/Rank.java
achillesecos: I ended up deleting this file. Did not need it since all the "Rank" functionality, comparator, is in Preferences.java

maceonthompson (approved):
LGTM!

src/main/java/com/google/sps/data/Preferences.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+
+import com.google.sps.data.Schedule;
+import com.google.sps.data.TimeRange;
+
+public class Preferences {
+    // Criteria sorted in order of user preference
+    private List<Preference> criteriaPreferences;
```
FabriceCastel, line 13: You might want to reconsider some of these names. If instead you had:

`Preference` -> `Criterion`

You could end up with something closer to..

```
public interface Criterion {
  public float score(Schedule schedule);
}

public class Preferences {
  private List<Criterion> criteria;

  ...
}

Criterion criterion = ...
float score = criterion.score(schedule);
```

That would avoid overloading the term "preference" in your code. It's honestly a fairly minor point though, feel free to take it or leave it.

src/main/java/com/google/sps/data/PrioritizeCoursesCriteria.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+  
+  /**
+   * Implements the preference score of the schedule, based on the criteria
+   * that prioritizes the order of courses that the user would like to take
+   */
+  public float preferenceScore(Schedule schedule) {
+    List<String> courseListNames = getCourseListNames(courseList);
+    int coursePrioritySum = 0;
+    List<Course> courses = (List) schedule.getCourses();
```
FabriceCastel: You should almost always be suspicious of any typecasting. Here it works because the `Collection<Course>` that `schedule.getCoutses()` returns happens to be a `List<Collection>` but you have no guarantee of this. If you really need a `List<T>`, you can get one like this:

```
List<Course> courses = new ArrayList(schedule.getCourses());
```

For this bit of code though, you don't really need a `List<T>` because `Collection<Course>` supports the foreach syntax, which is what you need it for:

```
Collection<Course> courses = schedule.getCourses();

for (Course course : courses) {
  ...
}
```

src/main/java/com/google/sps/data/PrioritizeCoursesCriteria.java
```diff
@@ -0,0 +1,41 @@
...
+    
+    for (Course course : courses) {
+      String courseName = course.getName();
+      int priorityIndex = courseListNames.size() - courseListNames.indexOf(courseName);
```
FabriceCastel: `courseListNames.indexOf(...)` is an O(N) operation you're going to repeat for every course you want to score.

It would be both more efficient and simple if instead, the code...
1. Created a `HashMap<String, Integer> courseScores` member in this class
2. Instantiated and filled in all of the values in this map in your constructor
3. Did a O(1) lookup in this map to find the course scores when adding them up

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Implements the preference score of the schedule, based on the criteria
+   * that prioritizes the TimeRanges that a user does not want class during
+   */
+  public float preferenceScore(Schedule schedule) {
+    // count the number of minutes that overlap
+    float minutesPenalty = 0;
+    
+    List<Course> courses = (List) schedule.getCourses();
```
FabriceCastel: See the other comment about typecasting

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return minutesPenalty;
+  }
+
+  // Calculates the overlapping minutes between two TimeRanges
+  int calculateDifference(TimeRange a, TimeRange b) {
+    int start = Math.max(a.start(), b.start());
+    int end = Math.min(a.end(), b.end());
+    return end - start;
+  }
```
FabriceCastel: This functionality seems like it belongs in `TimeRange` rather than here. You should consider instead implementing something like `TimeRange::calculateMinutesOverlap(TimeRange other)` (with unit tests for the edge cases, namely going over midnight, crossing over Sunday to Monday...)

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
```diff
@@ -0,0 +1,53 @@
...
+
+  // Returns a List of TimeRanges from a list of courses
+  List<TimeRange> getCourseTimes(List<Course> courses) {
+    List<TimeRange> times = new ArrayList<TimeRange>();
+    for(Course course : courses) {
+      times.addAll(course.getSections().get(0).getMeetingTimes());
```
FabriceCastel: `get(0)` ?

Can you assume this is always non-empty? Do you always want to ignore elements beyond the first?

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Implements the preference score of the schedule, based on the criteria
+   * that prioritizes the TimeRanges that a user does not want class during
+   */
+  public float preferenceScore(Schedule schedule) {
+    // count the number of minutes that overlap
+    float minutesPenalty = 0;
+    
+    List<Course> courses = (List) schedule.getCourses();
+    List<TimeRange> timeOrderedCourses = getCourseTimes(courses);
```
FabriceCastel: `timeOrderedCourses` -> `courseTimes`?

There's nothing here to indicate that `getCourseTimes()` returns a sorted list, nor that you need it sorted in this code.

src/main/java/com/google/sps/data/SubjectCoursesCriteria.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  /**
+   * Implements the preference score of the schedule, based on the criteria
+   * that prioritizes the number of courses pertaining to a particular subject
+   */
+  public float preferenceScore(Schedule schedule) {
+    int preferredSubjectCount = 0;
+    List<Course> courses = (List) schedule.getCourses();
+    for (Course course : courses) {
```
FabriceCastel, line 24: nit: it's considered more legible to use a linebreak before and after code blocks if (and only if) you have other lines of code before and/or after it.

Here are all four possible cases and how they should be broken out with newlines:

```
int foo() {
  while (something) {
    doSomethingElse();
  }
}

int foo() {
  int another = lineOfCode;

  while (something) {
    doSomethingElse();
  }
}

int foo() {
  while (something) {
    doSomethingElse();
  }

  int another = lineOfCode;
}

int foo() {
  int first = lineOfCode;

  while (something) {
    doSomethingElse();
  }

  int last = lineOfCode;
}
```

This applies for any nested code blocks, so... `if/else`, `switch`, `while` and so on would most often be written like this.

FabriceCastel (changes requested):

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
maceonthompson: This is the way that the `Scheduler::generateSchedules()` function outputs it's schedules. Each course in the outputted schedule has only one section (in accordance to the invariants, which state that no schedule can have multiple course sections in it).

achillesecos: As Maceo mentioned, each course in a schedule will only have one section (an invariant), so the course section is always the first element of the sections list. Perhaps I can discuss with Maceo on a more clear approach, i.e. `getSection()` that only returns that section, and avoids the confusing `getSections().get(0)`

src/main/java/com/google/sps/data/PrioritizeCoursesCriteria.java
achillesecos: Thanks for point that out. It seems that I don't need the List<T>, so I changed it to Collection<T>

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
FabriceCastel: Ok. I left a comment about this from the `Scheduler::generateSchedules()` point of view on another pull request here: https://github.com/googleinterns/step78-2020/pull/15#discussion_r451746758

Let me know how you would like to resolve this between both pull requests.

The easiest option, if you decide to go with the approach suggested in the linked comment, would be to open up a separate pull request to add `CourseEnrollment` and update the `Scheduler` API accordingly (only the return types and minimum required changes to make the types check out). Get that approved and merged quickly, then update both this and PR #15 accordingly.

...or you can do it gladiator style and see who gets to merge first then deal with the ensuing merge conflicts & breakages. Let me know what you settle on.

maceonthompson: See https://github.com/googleinterns/step78-2020/pull/15#discussion_r451785869

achillesecos: We decided on creating the `ScheduledCourse` class, representing a course that's added to the schedule with a single section rather than a `List<Section>`

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
achillesecos: Changed to `courseTimes`. Thanks for the catch, meant to change this.

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
achillesecos: Same as before, also changed to Collection.

src/main/java/com/google/sps/data/SubjectCoursesCriteria.java
achillesecos: Sounds good. I made the changes to the 3 criteria files

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
achillesecos: Sounds good. I added this to `TimeRange.java` and added unit tests in `TimeRangeTests.java`

src/main/java/com/google/sps/data/PrioritizeCoursesCriteria.java
achillesecos: That's a good idea. Only thing I changed was defining HashMap with <Course, Integer>

src/main/java/com/google/sps/data/Preferences.java
achillesecos: Thanks for the suggestion. I will probably do this in a different PR later

src/main/java/com/google/sps/data/RestrictTimesCriteria.java
achillesecos: I will make the changes to change `getSections().get(0)` once [PR  #21](https://github.com/googleinterns/step78-2020/pull/21) gets merged in.

maceonthompson: 21 is merged!

src/test/java/com/google/sps/TimeRangeTest.java
```diff
@@ -0,0 +1,85 @@
...
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void sunToMonOverlapTest() {
+    int overlap = timeRangeF.calculateMinutesOverlap(timeRangeG);
+    int actual = overlap;
+    int expected = DURATION_1_HOUR;
+    Assert.assertEquals(expected, actual);
+  }
```
FabriceCastel, line 104: Does this test pass? It looks like the implementation might calculate a large negative overlap and return zero here instead of one hour.

FabriceCastel (changes requested):
There's only one comment left, ~~plus the `.get(0)` fix to address~~ before this is good to go.

src/test/java/com/google/sps/TimeRangeTest.java
achillesecos: Yes `sunToMon` works. I realized what the issue was; the week "resets" from Sat -> Sun, so I created tests `satToSun` and it returns exactly what you describe. I added that edgecase check to the function so that it handles it [here](https://github.com/googleinterns/step78-2020/pull/14/commits/40cac5c14f07a96b94b1c0904b1c99c075a84ae6)

FabriceCastel (approved):
LGTM

PR #15, opened 2020-07-06
Add Schedule Permutation

maceonthompson:
~Contingent on #13 being merged, as `sectionsOverlap()` (line 126, Scheduler.java) relies heavily on the `Section.overlaps()` method~
13 has been merged, this is now ready to merge pending review


src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -85,14 +86,143 @@ public void notEnoughCreditsNoSchedules() {
...
+
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void requiredCoursesOverlapReturnsEmpty() {
+    Section section1 = new Section("Mario", monWedFri(10, 0, DURATION_60_MINUTES));
+    Section section2 = new Section("Sonic", monWedFri(10, 30, DURATION_60_MINUTES));
+    Section section = new Section("Dr. Eggman", tuesThurs(13, 30, DURATION_30_MINUTES));
+
```
hjohns28, line 173: minor nit: extra line of space here (and on function below)

hjohns28 (approved):
LGTM

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
-        return null;
+  /**
+   * Generates all valid schedules given a list of courses and invariant
+   * constraints.
+   * 
+   * @param courses    The list of courses to generate the schedules from.
+   * @param invariants The invariants that constrain the valid schedules
+   * @return A list of valid schedules.
+   */
+  public List<Schedule> generateSchedules(List<Course> courses, Invariants invariants) {
```
FabriceCastel, line 25: This is more of an FYI, not even a nit really. I should've thought to mention this earlier.

"Invariant" refers to a property that remains true after applying some transformation. This is typically used when establishing proofs that an algorithm works, where you'll describe some property as invariant when it should be true both before and after a function call/iteration of a loop/etc.

Here, the term "constraints" would work best, since it describes the set of conditions that the solution must abide by.

Again, not something you need to act on. Very niche nit.

src/test/java/com/google/sps/SchedulerTest.java
maceonthompson: Thanks!

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+    // Generate all possible valid Schedules that include only required courses.
+    float reqCredits = totalCredits(requiredCourses);
+    Set<List<Course>> requiredSet = new HashSet<>();
+
+    generateSchedulesHelper(requiredCourses, new ArrayList<>(), 
+        new Invariants(reqCredits, reqCredits), requiredSet);
+    
+    // Using those Schedules as a building block, generate all valid schedules
+    Set<List<Course>> scheduleSet = new HashSet<>();
+    for (List<Course> validCourseList : requiredSet) {
```
FabriceCastel: nit: `validCourseList` -> `requiredCourseList` ?

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+  /**
+   * Recursively iterates through a list of courses and adds schedules to the
+   * given set.
+   * 
+   * @param courses     The list of courses to iterate through
+   * @param courseList  The current course list to be added/operated on
+   * @param invariants  The invariants that constrain the valid schedules
+   * @param courseLists The set of course lists that is being generated/added to.
+   */
+  private void generateSchedulesHelper(List<Course> courses, List<Course> courseList,
```
FabriceCastel: The choice of param names `courses` and `courseList` makes it difficult to distinguish what does what. The comment at the top also leaves this ambiguous. Without reading the code itself, all I know is that both of these are lists of courses you plan to use in some way. Some potentially more descriptive name pairs:

```
unscheduledCourses / scheduledCourses
availableCourses / selectedCourses
```

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+   * Recursively iterates through a list of courses and adds schedules to the
+   * given set.
+   * 
+   * @param courses     The list of courses to iterate through
+   * @param courseList  The current course list to be added/operated on
+   * @param invariants  The invariants that constrain the valid schedules
+   * @param courseLists The set of course lists that is being generated/added to.
+   */
+  private void generateSchedulesHelper(List<Course> courses, List<Course> courseList, 
+      Invariants invariants, Set<List<Course>> courseLists) {
```
FabriceCastel: Similar to the comment above, `courseLists` is hard to understand here. Alternative: `generatedCourseLists`. It's not that great of a suggestion but seems like a step in the right direction.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+
+    // This means that we have a valid schedule, and should add it to the list.
+    float currCredits = totalCredits(courseList);
+    if (invariants.meetsCreditRequirement(currCredits)) {
+      courseLists.add(courseList);
+    }
+
+    // We've reached the end of the list or have a full schedule, we are finished.
+    if (courses.isEmpty() || currCredits >= invariants.getMaxCredits()) {
```
FabriceCastel: The max course credits number is treated as exclusive here. Was this a deliberate design decision?

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+      return;
+    }
+
+    Course course = courses.remove(0);
+    List<Section> courseSections = course.getSections();
+    List<Section> scheduleSections = courseList.stream()
+        .map(c -> c.getSections().get(0))
```
FabriceCastel: This relies on the assumption that each course has exactly one section.

If you want to get the list of all schedule sections, including the courses that have zero sections or more than one sections, you can use `Stream::flatMap()` instead:

```
    List<Section> scheduleSections = courseList.stream()
        .flatMap(c -> c.getSections().stream())
        .collect(Collectors.toList());
```

This may or may not be what you want though. Depending on how you handle this, you could potentially end up trying to add two sections of the same course to a schedule.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+     */
+    // TODO: Implement logic for required courses having no valid schedules
+    for (Section section : courseSections) {
+      boolean doesNotOverlapSchedule = !sectionsOverlap(section, scheduleSections);
+      boolean doesNotExceedCreditLimit = 
+          currCredits + course.getCredits() <= invariants.getMaxCredits();
+      
+      if (doesNotOverlapSchedule && doesNotExceedCreditLimit) {
+        List<Course> newCourseList = new ArrayList<>(courseList);
+        newCourseList.add(new Course(course.getName(), course.getCourseID(), course.getSubject(),
```
FabriceCastel: It seems like you're using the `Course` class in two slightly different ways:
1. To represent a course offering with multiple available sections
2. To represent a specific course section a student is enrolled in

It would be best for these two things to either be well represented with separate data structures, or represented by data structures that compose differently to represent both.

One approach that requires a less substantial change could be to use a different structure to represent course enrollment:

```
class CourseEnrollment {
  public Course getCourse();
  public Section getEnrolledSection();
}
```

It's not necessarily the cleanest solution here, but it works well enough and avoids this mild misuse of the `Course` class.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+    if (!course.isRequired()) {
+      generateSchedulesHelper(new ArrayList<>(courses), courseList, invariants, courseLists);
+    }
+  }
+
+  /**
+   * Returns the total number of credits in a given list of courses.
+   */
+  private float totalCredits(List<Course> courseList) {
+    float total = 0;
```
FabriceCastel, line 126: FYI - no action needed

Using `float` when you need to check for hard boundaries can cause you some problems due to the imprecise nature of floating point arithmetic. If you're curious about this, you can look into how financial systems/banks/etc handle currency calculations. The safest thing to do, if you're not going to have an arbitrarily small unit to work with, is to use ints and scale things up enough to guarantee whole numbers. In this case, if the smallest possible course credit number is 0.5 or 0.25, you could scale up all credits by a factor of 100 and use integers instead.

Here, you're *probably* going to be ok. In general this is a helpful thing to keep in mind though.

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -1,15 +1,134 @@
...
+   */
+  private boolean sectionsOverlap(Section toCheck, List<Section> sections) {
+    for (Section section : sections) {
+      if (toCheck.overlaps(section)) {
+        return true;
+      }
     }
-}
\ No newline at end of file
+    return false;
```
FabriceCastel, line 143: nit: add newlines before and after code blocks if (and only if) there are other lines of code before/after it.

Refer to this comment for the verbose version: https://github.com/googleinterns/step78-2020/pull/14#discussion_r451203461

FabriceCastel (changes requested):

src/main/java/com/google/sps/Scheduler.java
maceonthompson: We're creating a new class for courses that have been scheduled (called ScheduledCourse). It's essentially the same as the original Course class, but the `sections` field is now a single `section` as opposed to a list.

src/main/java/com/google/sps/Scheduler.java
maceonthompson: No, this is a bug. Thanks

src/main/java/com/google/sps/Scheduler.java
maceonthompson: Resolved!

FabriceCastel:
There's a lot of commits showing up here that have already been merged into master (eg. I can see the google API additions in the pom.xml and those are already in master). Can you either rebase this beanch onto master or merge master into this branch to clean up overlapping code changes?

FabriceCastel:
By the by, this is a prime example of the sort of pull request where you would want to use the "squash and merge" strategy once it gets reviewed.

maceonthompson:
> There's a lot of commits showing up here that have already been merged into master (eg. I can see the google API additions in the pom.xml and those are already in master). Can you either rebase this beanch onto master or merge master into this branch to clean up overlapping code changes?

I meant to ask this - those commits that are showing up happened specifically after I rebased this branch onto master. What would be the best way to fix that (how would I merge master into this branch for example) and how can I avoid that going forward?

kristen-law:
For reviewing purposes, here's a diff of this branch with master: https://github.com/googleinterns/step78-2020/compare/master...Permutation

maceonthompson:
Update: going to just squash and merge when approved, not going to deal w rebasing because it's too much of a headache (though will definitely be a lot more cognizant in the future wrt rebasing vs merging)

FabriceCastel (approved):
LGTM

+1 for that diff view @kristen-law

PR #16, opened 2020-07-07
display schedules (currently just one-hard-coded schedule) on secondary Google calendar

hjohns28:
notes: 
-- There is an issue with the timezone, that I am working on (it doesn't seem to be registering the timezone when I set it)
-- I just have a single-course hard-coded "schedule" right now, that I mostly borrowed from one of Maceo's tests, but I am going to add to / change that 
-- I added a function to Schedule.java that returns the courses, so that I could iterate through them when putting the schedule on the calendar
-- Right now the user just click's the single button (after they are logged in), and then it displays the one calendar


src/main/java/com/google/sps/data/TesterSchedule.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Returns a list of times on Tuesday and Thursday, to be used in section constructors.
+   */
+  public List<TimeRange> tuesThurs(int hour, int minute, int durationMinutes) {
+    TimeRange tues = TimeRange.fromStartDuration(TimeRange.TUESDAY, hour, minute, durationMinutes);
+    TimeRange thurs = TimeRange.fromStartDuration(TimeRange.THURSDAY, hour, minute, durationMinutes);
+    return Arrays.asList(tues, thurs);
+  }
+
```
maceonthompson, line 49: Not a requested change, but worth adding tests with a schedule that has multiple courses in it.

src/main/java/com/google/sps/server/AddSchedule.java
```diff
@@ -0,0 +1,212 @@
...
+    int day = dayOfWeek + nextSun;
+    String dayString = "0" + Integer.toString(day);
+    if (day > 9) {
+      dayString = Integer.toString(day);
+    }
+
+    String monthString = "0" + Integer.toString(month);
+    if (month > 9) {
+      monthString = Integer.toString(month);
+    }
```
maceonthompson: Not a necessary change, but a cool little "pretty" thing that you could do to make these more compact. Java's Ternary operator would be useful here, something like 
`String monthString = month > 9 ? Integer.toString(month) : "0" + Integer.toString(month)`
or 
```
String monthString = month > 9 
  ? Integer.toString(month) 
  : "0" + Integer.toString(month)
```
(These are the same thing, just different formatting)

maceonthompson (approved):
LGTM!

src/main/java/com/google/sps/data/TesterSchedule.java
hjohns28: Yes, I was planning on doing that now! Thank you

src/main/java/com/google/sps/server/AddSchedule.java
hjohns28: Thanks for the suggestion, that's definitely nicer than what I have, I'll change that.

src/main/java/com/google/sps/server/AddSchedule.java
```diff
@@ -0,0 +1,197 @@
...
+   * Puts schedule onto secondary calendar
+   * 
+   * @param client the user's Google calendar 
+   * @param calendarId the id of the secondary calendar to put the schedule on
+   */
+  public static void addSchedule(Schedule schedule, com.google.api.services.calendar.Calendar client, String calendarId) throws IOException {
+    try {
+      //get current date, so can add the courses to the appropriate days of the current month 
+      java.util.Calendar calendar = java.util.Calendar.getInstance();
+      int nextSun = calculateDateOfNextSunday(calendar);
```
achillesecos: Small suggestion: can maybe rename `calculateDateOfNextSunday()` to `getNextSundayDate()` to be inline with the next 2 function names

src/main/java/com/google/sps/server/AddSchedule.java
```diff
@@ -0,0 +1,197 @@
...
+      int month = getCurrentMonth(calendar);
+      int year = getCurrentYear(calendar);
+
+      //for each course in the schedule, add an event to the calendar
```
achillesecos: Nit: space between // and for

src/main/java/com/google/sps/data/TesterSchedule.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.ArrayList;
+
+public final class TesterSchedule {
+
+  private static final int DURATION_30_MINUTES = 30;
+  private static final int DURATION_60_MINUTES = 60;
+  private static final int DURATION_90_MINUTES = 90;
+  private static final int DURATION_1_HOUR = 60;
+  private static final int DURATION_2_HOUR = 120;
```
kristen-law: Do you need all of these? It looks like you're only using DURATION_90_MINUTES.

src/main/java/com/google/sps/server/AddSchedule.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Date;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Collection;
+import java.util.Iterator;
+import java.io.IOException;
+
+import com.google.sps.data.*;
+
+class AddSchedule {
```
kristen-law: Class names should generally be nouns rather than verbs because they typically represent objects. In this case, since it seems like all of the methods are calendar-related utilities, I'd recommend a name like `CalendarUtils`.

The `addSchedule` and `addEvent` methods seem like they should be instance methods of some `ScheduleCalendar` class rather than static methods in this class. This would allow you to do something like `calendar.addEvent` to directly add an event to some `calendar` object.

src/main/java/com/google/sps/server/Auth.java
```diff
@@ -19,50 +19,34 @@
...
+
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
```
kristen-law, line 25: Double-check these imports and remove the ones that are no longer being used in this file.

src/main/java/com/google/sps/server/HandleSchedules.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Exports the ordered schedules to secondary calendars on the user's Google calendar    
+ */
+@WebServlet("/handleSchedules")
+public class HandleSchedules extends HttpServlet {
+  
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      com.google.api.services.calendar.Calendar client = Utils.loadCalendarClient();
```
kristen-law: Add `com.google.api.services.calendar.Calendar` to your imports since it's the only `Calendar` class in this file.

src/main/java/com/google/sps/server/CreateCalendar.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.api.services.calendar.model.CalendarList;
+import com.google.api.services.calendar.model.CalendarListEntry;
+
+import java.util.Date;
+import java.util.ArrayList;
+import java.util.List;
+
+import java.io.IOException;
+
+class CreateCalendar {
```
kristen-law: Similar to my other comment, class names should generally not be verbs. For this case specifically, this `createCalendar` method doesn't need to be in its own class. It could instead be moved to a `ScheduleCalendar` class and used in the constructor.

src/main/java/com/google/sps/server/AddSchedule.java
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param timeInMin the time, in minutes of the week, for an event (either the start or end time)
+   * @param month the current month
+   * @param year the current year
+   * @param nextSun the day of the month of the next Sunday
+   * 
+   * @return a String formatted for DateTime, to be used in creating a calendar event 
+   */
+  public static String 
+     DateTime(int timeInMin, int month, int year, int nextSun) {
```
kristen-law: Is this supposed to be your `calculateDateTime` method?

src/main/java/com/google/sps/server/HandleSchedules.java
```diff
@@ -0,0 +1,67 @@
...
+      Collection<Schedule> schedules = new ArrayList<>();
+      schedules.add(schedule);
+
+      String calendarId = null;
+
+      //for each schedule, create new secondary calendary and put schedule on it
+      Iterator<Schedule> iterator = schedules.iterator();
+      while (iterator.hasNext()) {
+        calendarId = CreateCalendar.createCalendar(client);
+        AddSchedule.addSchedule(iterator.next(), client, calendarId);
```
kristen-law: Instead of creating separate classes `CreateCalendar` and `AddSchedule` with static methods, it might make more sense from an object-oriented programming perspective to create a new class that represents the secondary calendar named `ScheduleCalendar` or something similar. You would be able to store variables like the `calendarId` and have methods like `addSchedule` directly within that class.

kristen-law (changes requested):

src/main/java/com/google/sps/server/Auth.java
hjohns28: Thanks for the reminder. I've been checking on the new files, but (obviously) missed some on the old files, sorry about that!

src/main/java/com/google/sps/server/HandleSchedules.java
hjohns28: This does make more sense, thank you for the suggestion. I'll make that change!

src/main/java/com/google/sps/server/AddSchedule.java
hjohns28: Yes, in my code it says calculateDateTime, not sure what happened but it should be fixed when I commit the changes for this review. My bad!

src/main/java/com/google/sps/server/ScheduleDisplayHandler.java
```diff
@@ -0,0 +1,68 @@
...
+
+      List<String> calIds = new ArrayList<>();
+
+      //for each schedule, create new secondary calendary and put schedule on it
+      Iterator<Schedule> iterator = schedules.iterator();
+      while (iterator.hasNext()) {
+        ScheduleCalendar cal = new ScheduleCalendar(client);
+        cal.addSchedule(iterator.next());
+        calIds.add(cal.getCalendarId());
+      }
```
kristen-law: While this works, I think it might improve code readability to use a for-each loop to iterate over `schedules` instead of using an iterator. This might be more of a personal preference though, so up to you if you want to change it.

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+   /**
+   * Helper function to calculate the day of the month of the next Sunday,
+   * to know when to start the schedule. 
+   *
+   * @param calendar an instance of Calendar, so can know the current date
+   * @return an int representing the date of next Sunday
+   */
+  public static int getDateOfNextSunday(java.util.Calendar calendar) {
```
kristen-law: All of your static helper functions in this class can be moved to a separate CalendarUtils class since they're not dependent on having an instance of `ScheduleCalendar`. It's not 100% necessary, but since this file is getting a bit long it will help with code readability to have just the instance methods here: `getCalendarId`, `addEvent`, and `addSchedule`.

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -0,0 +1,250 @@
...
+      Iterator<Course> courseIterator = schedule.getCourses().iterator();
+      while (courseIterator.hasNext()) {
+        Course currentCourse = courseIterator.next();
+        List<TimeRange> sectionTimes = currentCourse.getSections().get(0).getMeetingTimes();
+        for (int i = 0; i < sectionTimes.size(); i++) {
+          String startTime = calculateStartTime(sectionTimes.get(i), month, year, nextSun);
+          String endTime = calculateEndTime(sectionTimes.get(i), month, year, nextSun);
+          addEvent(currentCourse, startTime, endTime);
+        }
+      }
```
kristen-law: Same comment on using for-each instead of iterator here to improve code readability.

kristen-law (changes requested):
Looks good, just a couple minor code readability comments.

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -0,0 +1,250 @@
...
+   /**
+   * Helper function to add a course in the schedule to the calendar, as an event. 
+   *
+   * @param course the course being added 
+   * @param calendarId the id of the secondary calendar to add the course to
+   * @param startTime the start time of the course
+   * @param endTime the end time of the course
+   */
+
+  public void addEvent(Course course, String startTime, String endTime) throws IOException {
```
kristen-law: It looks like this is only being used internally within this class, so this should be a private method.

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -71,18 +68,16 @@ public void addSchedule(Schedule schedule) throws IOException {
...
-      int year = getCurrentYear(calendar);
+      int nextSun = CalendarUtils.getDateOfNextSunday(calendar);
+      int month = CalendarUtils.getCurrentMonth(calendar);
+      int year = CalendarUtils.getCurrentYear(calendar);
 
       // for each course in the schedule, add an event to the calendar
-      Iterator<Course> courseIterator = schedule.getCourses().iterator();
-      while (courseIterator.hasNext()) {
-        Course currentCourse = courseIterator.next();
+      for (Course currentCourse: schedule.getCourses()) {
```
kristen-law: Nit: there should be a space between `currentCourse` and the colon

src/main/java/com/google/sps/server/ScheduleDisplayHandler.java
```diff
@@ -51,10 +50,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       List<String> calIds = new ArrayList<>();
 
       //for each schedule, create new secondary calendary and put schedule on it
-      Iterator<Schedule> iterator = schedules.iterator();
-      while (iterator.hasNext()) {
+      for (Schedule currentSchedule: schedules) {
```
kristen-law: Nit: add a space between `currentSchedule` and colon

kristen-law (approved):
LGTM - Just a small style fix needed, then feel free to merge

PR #18, opened 2020-07-08
Scheduled course

maceonthompson:
Add a "ScheduledCourse" class to represent courses that have already been added to a schedule.


achillesecos (approved):
LGTM

FabriceCastel (approved):
LGTM - you might want to do `ScheduledCourse extends Course` sometime in the future to avoid code duplication since this is almost a strict superset of that class.

PR #20, opened 2020-07-08
Delete secondary calendars not chosen by user 

hjohns28:


maceonthompson (approved):
LGTM

src/main/java/com/google/sps/server/DeleteCalendar.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Deletes secondary calendar 
+ */
+@WebServlet("/deleteCalendar")
+public class DeleteCalendar extends HttpServlet {
+  
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      com.google.api.services.calendar.Calendar client = Utils.loadCalendarClient();
```
kristen-law: This can just be `Calendar` since you're importing it already

src/main/webapp/script.js
```diff
@@ -53,3 +54,14 @@ function displayCalendar(calId) {
   const url = "https://calendar.google.com/calendar/embed?src=" + calId;
   document.getElementById("calendar").src = url;
 }
+
+// delete all secondary calendars created, except for the chosen one 
+function deleteUnchosenCalendars() {
+  for (var j = 0; j < idArray.length; j++) {
+    if (j != i) {
```
kristen-law: I'd recommend renaming `i` to something like `currCalendarIndex` to be a bit more descriptive since it's used across different functions.

achillesecos (approved):
LGTM

kristen-law (changes requested):

src/main/webapp/script.js
```diff
@@ -53,3 +54,14 @@ function displayCalendar(calId) {
   const url = "https://calendar.google.com/calendar/embed?src=" + calId;
   document.getElementById("calendar").src = url;
 }
+
+// delete all secondary calendars created, except for the chosen one 
+function deleteUnchosenCalendars() {
```
FabriceCastel, line 59: Slightly nicer/more legible way to write this:

```
function deleteUnchosenCalendars() {
  idArray
    .filter((id, index) => index !== chosenCalendarIndex)
    .forEach(id => deleteCalendar(id));

  const chosenCalendarId = idArray[chosenCalendarIndex];
  idArray = [chosenCalendarId];

  displayCalendar(chosenCalendarId);
}

function deleteCalendar(int calendarId) {
  fetch("/deleteCalendar?calId=" + id, { method: 'POST' });
}
```

hjohns28: This is nicer, thanks for showing me! I'll change my code.

src/main/webapp/script.js
```diff
@@ -53,3 +54,19 @@ function displayCalendar(calId) {
...
+  idArray
+      .filter((id, index) => index !== currCalendarIndex)
+      .forEach(id => deleteCalendar(id));
+
+  const chosenCalendarId = idArray[currCalendarIndex];
+  idArray = [chosenCalendarId];
```
kristen-law: You'll probably want to set `currCalendarIndex` to 0 here also.

kristen-law (approved):
LGTM - just one small comment to address before merging

src/main/webapp/script.js
hjohns28: Good point, I'll make that change.

PR #21, opened 2020-07-08
adjust ScheduledCourse to extend Course

maceonthompson:
I don't know if I actually needed to implement the hashcode fucnction, as it just calls super.hashcode

### DIFF
--- a/src/main/java/com/google/sps/data/ScheduledCourse.java
+++ b/src/main/java/com/google/sps/data/ScheduledCourse.java
@@ -50,7 +50,7 @@ public class ScheduledCourse extends Course {
 
   @Override
   public int hashCode() {
-    return super.hashCode();
+    return super.hashCode() * Objects.hashCode(section);
   }
 
   public Section getSection() {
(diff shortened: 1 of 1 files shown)


hjohns28 (approved):
changes LGTM

src/main/java/com/google/sps/data/ScheduledCourse.java
```diff
@@ -38,60 +35,26 @@
...
-        && otherCourse.isRequired == this.isRequired 
-        && otherCourse.section.equals(this.section);
+    return false;
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(name, courseID, subject, credits, isRequired, section);
+    return super.hashCode();
+  }
```
kristen-law: You're right, you don't need to override `hashCode()` if you're just calling `super.hashCode()` since it will inherit that method. This can be removed.

I'm curious though since you adjusted the `equals()` method to only allow equality between two ScheduledCourses and not a ScheduledCourse and a Course, is it ok for a ScheduledCourse and a Course to hash to the same value?

kristen-law (approved):
LGTM, just one comment

src/main/java/com/google/sps/data/ScheduledCourse.java
maceonthompson: You're right! Ideally they wont. Addressed this issue.

PR #29, opened 2020-07-13
fix login by adding welcome page 

hjohns28:
the basic new setup: 
--index.html is now the welcome page that will explain our app, with only one way out, which sends the user to auth to login 
--once logged in, they'll be redirected to course-form.html (this currently contains what used to be in index.html, and displays the schedules, but will soon contain the input form for courses)


maceonthompson (approved):
LGTM!

FabriceCastel (approved):
LGTM

PR #32, opened 2020-07-15
update calendar display to add events for lab times, if applicable

hjohns28:


maceonthompson (approved):
LGTM!

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -74,12 +74,20 @@ public void addSchedule(Schedule schedule) throws IOException {
 
       // for each course in the schedule, add an event to the calendar
       for (ScheduledCourse currentCourse : schedule.getCourses()) {
-        List<TimeRange> sectionTimes = currentCourse.getSection().getMeetingTimes();
-        for (int i = 0; i < sectionTimes.size(); i++) {
-          String startTime = CalendarUtils.calculateStartTime(sectionTimes.get(i), month, year, nextSun);
-          String endTime = CalendarUtils.calculateEndTime(sectionTimes.get(i), month, year, nextSun);
+        List<TimeRange> lectureSectionTimes = currentCourse.getLectureSection().getMeetingTimes();
```
achillesecos, line 73: Is `getLectureSection()` implemented in ScheduledCourse, or are you filling it in for now

hjohns28: Maceo already added getLectureSection() to ScheduledCourse on the HandleRecitations branch

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -74,12 +74,20 @@ public void addSchedule(Schedule schedule) throws IOException {
...
+        for (int i = 0; i < lectureSectionTimes.size(); i++) {
+          String startTime = CalendarUtils.calculateStartTime(lectureSectionTimes.get(i), month, year, nextSun);
+          String endTime = CalendarUtils.calculateEndTime(lectureSectionTimes.get(i), month, year, nextSun);
           addEvent(currentCourse, startTime, endTime);
         }
```
kristen-law: It's best not to repeat the same code twice if you can avoid it. In this case, you could create a private helper function that adds a Section to the calendar.

src/main/java/com/google/sps/server/ScheduleCalendar.java
achillesecos: Oh ok. I see it now, I completely missed it then

hjohns28: No problem !

kristen-law (changes requested):

src/main/java/com/google/sps/server/ScheduleCalendar.java
hjohns28: Good point, I will make that improvement now, thanks!

hjohns28: Instead of adding a helper function that then called addEvent(), or returning both the startTime and endTime to addSchedule() and then calling addEvent(), I just added the two startTime / endTime calculations to addEvent(), because this still avoids the duplicate code and seemed simplest. If a helper function would still be clearer, though, let me know and I'll make the change!

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -107,12 +103,16 @@ public String getCalendarId() {
...
-   * @param startTime the start time of the course
-   * @param endTime the end time of the course
+   * @param sectionTime the timerange of the current course section
+   * @param month the current month, to be used in calculating the start and end times
+   * @param year the current year, to be used in calculating the start and end times
+   * @param nextSun the date of the next Sunday, to be used in calculating the start and end times
    */
-  private void addEvent(Course course, String startTime, String endTime) throws IOException {
+  private void addEvent(ScheduledCourse course, TimeRange sectionTime, int month, int year, int nextSun) throws IOException {
     try {
```
kristen-law, line 121: You don't necessarily need to change this now, but just a note that in general, you want to try to limit the code within the `try` block to only the code that might throw the exception.

src/main/java/com/google/sps/server/ScheduleCalendar.java
kristen-law: That solves the problem of the startTime and endTime being duplicated, but you still have two for-loops. It also makes the parameters being passed into `addEvent` a little more difficult to understand.

I'd recommend splitting out a separate function that contains the for-loop and calls `addEvent` with the previous parameters. Something like the following might work:
`private void addSectionToCalendar(Section section, java.util.Calendar calendar) { ... }`

Also something to consider, but don't necessarily need to change now: Your calculateStartTime and calculateEndTime in CalendarUtils could take a java.util.Calendar (instead of month, year, and nextSun) and then call the functions getCurrentMonth, getCurrentYear, and getDateOfNextSunday inside your start and end time functions. This would simplify things from a code readability perspective, since this class doesn't need to know the values of month, year, and nextSun.

hjohns28: That makes sense, that would be simpler. Thanks for the feedback, I'll work on that.

src/main/java/com/google/sps/server/ScheduleCalendar.java
hjohns28: sorry I didn't see this before re-requesting review, but thanks for letting me know. I'll keep this in mind and change it later if I have time!

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -62,30 +62,45 @@ public ScheduleCalendar(com.google.api.services.calendar.Calendar client) throws
...
+   */
+  private void addSectionToCalendar(List<TimeRange> sectionTimes, java.util.Calendar calendar, ScheduledCourse currentCourse) throws IOException {
+    try {
+      for (int i = 0; i < sectionTimes.size(); i++) {
+        String startTime = CalendarUtils.calculateStartTime(sectionTimes.get(i), calendar);
+        String endTime = CalendarUtils.calculateEndTime(sectionTimes.get(i), calendar);
+        addEvent(currentCourse, startTime, endTime);
+      }
+    } catch (IOException e) {
+      System.out.println("Couldn't iterate through section times");
```
kristen-law: Please update this error message, it's a little misleading. The exception isn't caused by not being able to iterate through the section times, but rather that the section couldn't be added to the calendar.

kristen-law (approved):
LGTM, feel free to submit after addressing the comment about the error message.

src/main/java/com/google/sps/server/ScheduleCalendar.java
hjohns28: Thanks for pointing that out, that's true.

PR #34, opened 2020-07-16
Scheduler recitation

maceonthompson:


src/main/java/com/google/sps/Scheduler.java
```diff
@@ -79,26 +79,50 @@ private void generateSchedulesHelper(List<Course> availableCourses, List<Schedul
...
+        boolean doesNotOverlapSchedule = !sectionsOverlap(lectureSection, scheduleSections);
+        boolean doesNotExceedCreditLimit = 
+            currCredits + course.getCredits() <= invariants.getMaxCredits();
+        
+        if (doesNotOverlapSchedule && doesNotExceedCreditLimit) {
+          List<ScheduledCourse> newScheduledCourseList = new ArrayList<>(currentSchedule);
+          newScheduledCourseList.add(new ScheduledCourse(course, lectureSection));
+          generateSchedulesHelper(new ArrayList<>(availableCourses), 
+              newScheduledCourseList, invariants, generatedScheduledCourseLists);
+        }
```
maceonthompson: This code is super redundant, but I can't think of a way to make it prettier. This is primarily because of the nested forLoop in the case where hasLabs is true (line 101).
I could move `doesNotExceedCreditLimit` outside of the larger if statement entirely, (as I write this I might do that as if that boolean is false looping over all the sections is completely unnecessary), but all of the other logic, despite being almost the same, is different enough that I don't know how I could abstract it further.

achillesecos: Maybe you can keep the outer loop, `for(Section lectureSection : courseLectureSection)` as is (meaning do not replicate) and put the `if(hasLabs)...else...` inside that outer loop, so that you don't need to replicate the outerloop. Im not completely certain it could work, i.e. scope of variables, but something worth considering

src/main/java/com/google/sps/data/ScheduledCourse.java
```diff
@@ -70,9 +87,17 @@ public ScheduledCourse(Course course, Section lectureSection) {
   public boolean equals(Object other) {
     if (other instanceof ScheduledCourse) {
       ScheduledCourse otherCourse = (ScheduledCourse) other;
```
achillesecos, line 89: If `other` is `ScheduledCourse` type, why do you cast it? Was this part of the issue you had before with equality?

src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -62,23 +63,41 @@ public void setUp() {
...
+    Section section = new Section("Dr. Eggman", monWedFri(10, 30, DURATION_90_MINUTES));
+    Course course1 = new Course("Intro to Evil", "EVIL100", 
+        "Wrongdoing", 1, true, Arrays.asList(section), Collections.emptyList());
+
+    ScheduledCourse t1 = new ScheduledCourse(course1, section);
+    ScheduledCourse t2 = new ScheduledCourse(course1, section, null);
+    ScheduledCourse t3 = new ScheduledCourse("Intro to Evil", "EVIL100", "Wrongdoing", 1, true, section);
+    ScheduledCourse t4 = new ScheduledCourse("Intro to Evil", "EVIL100", "Wrongdoing", 1, true, section, null);
+    ScheduledCourse[] courses = {t1, t2, t3, t4};
+    for(int i = 0; i < 4; i++) {
```
achillesecos: Maybe `courses.size()` instead of 4?

src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -233,4 +254,49 @@ public void requiredSectionsOverlapButWorks() {
     HashSet<Schedule> expected = new HashSet<>(Arrays.asList(schedule1, schedule2));
     Assert.assertEquals(expected, actual);
   }
+
+  @Test
+  public void LabLectureSectionsCountedAsOne()  {
```
achillesecos: Lowercase `lab` ?

src/main/java/com/google/sps/data/ScheduledCourse.java
maceonthompson: This is standard practice for custom .equals() methods - casting is the only way to access the other object's fields/methods. (If you look through the code at the other equals methods they cast as well)

src/test/java/com/google/sps/SchedulerTest.java
maceonthompson: Good catch, thanks!

src/main/java/com/google/sps/Scheduler.java
FabriceCastel: The indentation might be a bit off, but similar to what Achilles proposed:

```
for (Section lectureSection : courseLectureSections) {
  int totalCredits = currCredits + course.getCredits();
  boolean exceedsCreditLimit = totalCredits > invariants.getMaxCredits();

  if (exceedsCreditLimit) {
    continue;
  }

  for (Section labSection : courseLabSections) {
    boolean overlapSchedule = sectionsOverlap(
        lectureSection,
        labSection,
        scheduleSections);

    if (!overlapsSchedule) {
      ScheduledCourse scheduledCourse = new ScheduledCourse(course, lectureSection, labSection);

      List<ScheduledCourse> newScheduledCourseList = new ArrayList<>(currentSchedule);
      newScheduledCourseList.add(scheduledCourse);
      
      generateSchedulesHelper(
          new ArrayList<>(availableCourses), 
          newScheduledCourseList,
          invariants,
          generatedScheduledCourseLists);
    }
  }

  if (!hasLabSection) { 
    boolean overlapSchedule = sectionsOverlap(lectureSection, scheduleSections);

    if (!overlapSchedule) {
      ScheduledCourse scheduledCourse = new ScheduledCourse(course, lectureSection);

      List<ScheduledCourse> newScheduledCourseList = new ArrayList<>(currentSchedule);
      newScheduledCourseList.add(scheduledCourse);

      generateSchedulesHelper(
          new ArrayList<>(availableCourses), 
          newScheduledCourseList,
          invariants,
          generatedScheduledCourseLists);
    }
  }
}
```

You could clean this up further by abstracting out the overlap check and recursive call logic.

FabriceCastel (changes requested):

src/main/java/com/google/sps/Scheduler.java
maceonthompson: I didn't abstract out the overlap check/recursive call logic because the recursive call would require having the abstracted function take in all of the parameters for `generateSchedulesHelper` in addition to the parameters needed for the overlap check. Which would technically result in "cleaner" code, but a call to a helper-helper function with ~7 parameters felt uglier than just having 3-4 lines of redundant code.

FabriceCastel (approved):
LGTM

PR #35, opened 2020-07-22
Add user input conversion 

maceonthompson:
On submit, convert the state to a parseable JSON format for GSON on the backend.

### DIFF
--- a/frontend/src/InputForm.js
+++ b/frontend/src/InputForm.js
@@ -322,12 +322,12 @@ class InputForm extends React.Component {
 
   submit() {
     let submitState = JSON.parse(JSON.stringify(this.state));
-    console.log(JSON.stringify(this.state));
     submitState.courses = submitState.courses.map((course) => this.convertCourseSections(course));
 
     submitState.criterion.timePreferences = submitState.criterion.timePreferences.map((times) => 
         this.timeToTimeRange(0, times.startTime, times.endTime));
     var json = JSON.stringify(submitState);
+    //! This will be changed to send the json to the backend
     console.log(json);
   } 
 
--- a/frontend/src/InputForm.js
+++ b/frontend/src/InputForm.js
@@ -322,9 +322,11 @@ class InputForm extends React.Component {
 
   submit() {
     let submitState = JSON.parse(JSON.stringify(this.state));
-    console.log(submitState.courses);
+    console.log(JSON.stringify(this.state));
     submitState.courses = submitState.courses.map((course) => this.convertCourseSections(course));
 
+    submitState.criterion.timePreferences = submitState.criterion.timePreferences.map((times) => 
+        this.timeToTimeRange(0, times.startTime, times.endTime));
     var json = JSON.stringify(submitState);
     console.log(json);
   } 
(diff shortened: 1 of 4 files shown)


hjohns28 (approved):
LGTM

src/test/java/com/google/sps/JSONTest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonIOException;
+
+public final class JSONTest {
+  public static void main(String[] args) {
+    Gson gson = new Gson();
+    // JsonParser parser = new JsonParser();
```
kristen-law: Remove commented code

src/test/java/com/google/sps/JSONTest.java
```diff
@@ -0,0 +1,61 @@
...
+    try {
+      JsonObject wholeJSON = JsonParser.parseReader(new FileReader("/home/maceothompson/step78-2020/src/test/java/com/google/sps/everythingJSON.json")).getAsJsonObject();
```
kristen-law: You shouldn't be using an absolute file path here since it's specific to your own personal directory structure. You might need to use the file path relative to the project's root folder instead of relative to this file for this to work properly.

frontend/src/InputForm.js
```diff
@@ -320,10 +321,59 @@ class InputForm extends React.Component {
...
+      }
+      if (section.wednesday) {
+        newSection.meetingTimes.push(this.timeToTimeRange(3, section.startTime, section.endTime));
+      }
+      if (section.thursday) {
+        newSection.meetingTimes.push(this.timeToTimeRange(4, section.startTime, section.endTime));
+      }
+      if (section.friday) {
+        newSection.meetingTimes.push(this.timeToTimeRange(5, section.startTime, section.endTime));
+      }
```
kristen-law, line 351: You might want to consider storing the section days differently to avoid this duplicate code. For example, it could be an array of integers:

Sections on Tuesday and Thursday could be stored as section.days = [2, 4]

You could then loop over section.days and do this for each day in the array:
`newSection.meetingTimes.push(this.timeToTimeRange(<current element in array>, section.startTime, section.endTime));`

kristen-law (changes requested):

frontend/src/InputForm.js
maceonthompson: This would be ideal, but since the UI/user input states are one and the same, and the days of the week are checkboxes, the way that they're stored in state is as separate booleans. I'll check with Heidi to see if we can change the way that sections are stored, but for now this is the only option wrt: this function.

src/test/java/com/google/sps/JSONTest.java
maceonthompson: Resolved!

src/test/java/com/google/sps/JSONTest.java
maceonthompson: Resolved!

kristen-law (approved):

PR #36, opened 2020-07-22
add deletion of components

hjohns28:
(there are some more minor formatting things that aren't changed on this branch, but are fixed on userInput)

### DIFF
--- a/frontend/src/BasicInfo.js
+++ b/frontend/src/BasicInfo.js
@@ -36,12 +36,14 @@ class BasicInfo extends React.Component {
           minCredits={this.props.minCredits}
           maxCredits={this.props.maxCredits}
           updateMinCredits={this.updateMinCredits}
-          updateMaxCredits={this.updateMaxCredits}/>
+          updateMaxCredits={this.updateMaxCredits}
+        />
         <TermDates 
           startDate={this.props.startDate}
           endDate={this.props.endDate}
           updateTermStartDate={this.updateTermStartDate}
-          updateTermEndDate={this.updateTermEndDate}/>
+          updateTermEndDate={this.updateTermEndDate}
+        />
       </Card>
     )
   }
--- a/frontend/src/BasicInfo.js
+++ b/frontend/src/BasicInfo.js
@@ -3,46 +3,21 @@ import { Card } from '@material-ui/core';
 import Credits from './Credits';
 import TermDates from './TermDates';
 
-class BasicInfo extends React.Component {
-  constructor(props) {
-    super(props);
-
-    this.updateMinCredits = this.updateMinCredits.bind(this);
-    this.updateMaxCredits = this.updateMaxCredits.bind(this);
-    this.updateTermStartDate = this.updateTermStartDate.bind(this);
-    this.updateTermEndDate = this.updateTermEndDate.bind(this);
-  }
-
-  updateMinCredits(minCredits) {
-    this.props.updateMinCredits(minCredits);
-  }
-
-  updateMaxCredits(maxCredits) {
-    this.props.updateMaxCredits(maxCredits);
-  }
-
-  updateTermStartDate(startDate) {
-    this.props.updateTermStartDate(startDate);
-  }
-
-  updateTermEndDate(endDate) {
-    this.props.updateTermEndDate(endDate);
-  }
-  
+class BasicInfo extends React.Component {  
   render() {
     return (
       <Card>
         <Credits 
           minCredits={this.props.minCredits}
           maxCredits={this.props.maxCredits}
-          updateMinCredits={this.updateMinCredits}
-          updateMaxCredits={this.updateMaxCredits}
+          updateMinCredits={this.props.updateMinCredits}
+          updateMaxCredits={this.props.updateMaxCredits}
         />
         <TermDates 
           startDate={this.props.startDate}
           endDate={this.props.endDate}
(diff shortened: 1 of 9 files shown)


frontend/src/TimePreference.js
```diff
@@ -18,20 +21,29 @@ class TimePreference extends React.Component {
...
-          label="End time: " type="time"
-          defaultValue="17:00" onChange={this.handleEndTimeChange}
-          InputLabelProps={{ shrink: true, }} inputProps={{ step: 300, }}
-        />
-      </CardContent>
+      <Card>
+        <CardContent>
+          <TextField
+            label="Start time: " type="time"
+            defaultValue="08:00" onChange={this.handleStartTimeChange}
```
maceonthompson: One thing I've noticed (in very limited testing) is that this default value isn't applied, so if the user submits without actually changing the value it submits an empty string in the Start Time (or End Time) fields.
I don't think there's much one can do to change this aside from either removing the "defaultValue" (so the user knows that they need to change the input) or doing something fancier with state or something like that.
This isn't a big deal but I thought I'd mention it.

maceonthompson (approved):
LGTM!

frontend/src/TimePreference.js
hjohns28: Yes - thank you for bringing it up, I was going to mention that as well because I noticed it (and planned on changing it) too. I didn't want the user to think they didn't have to put in a time if it was the same as the default, so I was just going to remove it. I'll do that now since we've brought it up!

frontend/src/Course.js
```diff
@@ -86,21 +99,23 @@ class Course extends React.Component {
...
+          </IconButton>
           <Input placeholder="Subject" inputProps={{ 'aria-label': 'description' }}
             value={this.props.subject} onChange={this.handleSubjectChange} />
           <TextField placeholder="Credits" value={this.props.credits} onChange={this.handleCreditsChange}
             InputProps={{ inputComponent: NumberFormatCustom }} />
-          <FormControlLabel
-            control={<Switch checked={this.props.isRequired} onChange={this.handleIsRequiredChange} />}
-            label="Required"/>
-
+          <br></br>
```
FabriceCastel: `<br/>` ?

frontend/src/Criterion.js
```diff
@@ -23,23 +25,24 @@ class Criterion extends React.Component {
   updateSubjectPreference(subject) {
     this.props.updateSubjectPreference(subject);
   }
+
+  deleteTimePreference(id) {
+    this.props.deleteTimePreference(id);
+  }
```
FabriceCastel: I've seen this in a few places now. If your function is a straight pass through to a function in your component's props, you can refer to that prop without the additional level of indirection.

So, instead of this..

```
<SomeComponent deleteTimePreference={this.deleteTimePreference} ... />
```

You would do this..

```
<SomeComponent deleteTimePreference={this.props.deleteTimePreference} ... />
```

Without needing to define a `deleteTimePreference` method on your `Criterion` React component.

frontend/src/Criterion.js
```diff
@@ -23,23 +25,24 @@ class Criterion extends React.Component {
...
+            id={index}
+            key={index}
+            timePreference={timePreference}
+            updateTimeStartPreference={this.updateTimeStartPreference}
+            updateTimeEndPreference={this.updateTimeEndPreference}
+            deleteTimePreference={this.deleteTimePreference}/>))}
+          <Button onClick={this.props.createNewTimePreference}>+ time-block</Button>
           <SubjectPreference 
             subject={this.props.subject}
             updateSubjectPreference={this.updateSubjectPreference}/>
```
FabriceCastel: Please fix the indentation in this section

FabriceCastel (changes requested):

frontend/src/Criterion.js
hjohns28: I didn't realize that, thanks for pointing it out!

frontend/src/BasicInfo.js
```diff
@@ -3,46 +3,21 @@ import { Card } from '@material-ui/core';
...
-
-  updateTermStartDate(startDate) {
-    this.props.updateTermStartDate(startDate);
-  }
-
-  updateTermEndDate(endDate) {
-    this.props.updateTermEndDate(endDate);
-  }
-  
+class BasicInfo extends React.Component {
```
FabriceCastel: This component is functionally pure, so you can replace it with a lambda.

https://www.tutorialspoint.com/pure-component-in-react-js

frontend/src/Criterion.js
```diff
@@ -4,49 +4,24 @@ import Button from '@material-ui/core/Button';
...
-  
-  updateSubjectPreference(subject) {
-    this.props.updateSubjectPreference(subject);
-  }
-
-  deleteTimePreference(id) {
-    this.props.deleteTimePreference(id);
-  }
-  
+class Criterion extends React.Component {
```
FabriceCastel: Same comment as above about making this a pure component

FabriceCastel (changes requested):

frontend/src/BasicInfo.js
hjohns28: Got it, thanks for showing me this!

FabriceCastel (approved):
LGTM

PR #37, opened 2020-07-23
Add basic backend processing for user input

maceonthompson:
Currently only permutes schedules, does not handle sorting.


hjohns28 (approved):
LGTM

frontend/src/InputForm.js
```diff
@@ -364,20 +364,23 @@ class InputForm extends React.Component {
...
-    var json = JSON.stringify(submitState);
-    //! This will be changed to send the json to the backend
-    console.log(json);
-  } 
+    fetch("/handleUserInput", {
+      method:"POST",
+      headers: {'Content-Type':'application/json'},
+      body: JSON.stringify(submitState)
+    }).then(response => response.json())
+      .then(scheduleList => console.log(JSON.stringify(scheduleList)));
```
FabriceCastel: Should the `scheduleList` be stored anywhere so the frontend can read from it/render the schedules? This looks like placeholder code at the moment, but storing the variable at this point would still be better than logging it.

src/main/java/com/google/sps/server/ScheduleCreationHandler.java
```diff
@@ -14,64 +14,58 @@
...
-   * Helper function to instantiate hard-coded courses, and their preferred order 
-   *
-   * @return a List of the ordered Course objects      
-   */
-  private static List<Course> createCoursesPriority() {
-    Section section1 = new Section("Professor A", TesterSchedule.monWedFri(9, 00, DURATION_90_MINUTES));
-    Course course1 = new Course("Operating System", "15410", 
-        "Computer Science", 15, true, Arrays.asList(section1));
+    Type courseListType = new TypeToken<List<Course>>(){}.getType();
+    List<Course> courses = gson.fromJson(wholeJSON.getAsJsonArray("courses"), courseListType);
```
FabriceCastel: It would be cleaner if you had a single `GenerateScheduleRequest` class that contained all of the fields you care about (courses, basicInfo, credits...). You could then do:

```
GenerateScheduleRequest request = gson.fromJson(wholeJSON, GenerateScheduleRequest.class);
List<Course> courses = request.getCourses();
Invariants invariants = request.getCredits();
...
```

Something like that. One single structure to deserialize and then everything is normal type-checked Java from there on out.

It's much easier to work with a single class definition that you then serialize to and from.

FabriceCastel (changes requested):

FabriceCastel (approved):
LGTM

PR #38, opened 2020-07-23
Fix bug where non-Required courses weren't being properly added to schedules

maceonthompson:
As a fun side effect, `scheduler.generateSchedulesHelper` is now more space efficient

### DIFF
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -59,10 +59,11 @@ public final class Scheduler {
    * Recursively iterates through a list of courses and adds schedules to the
    * given set.
    * 
-   * @param courses     The list of courses to iterate through
-   * @param currentSchedule  The current scheduled course list to be added/operated on
-   * @param invariants  The invariants that constrain the valid schedules
-   * @param courseLists The set of course lists that is being generated/added to.
+   * @param index                         The index the current course in availableCourses
+   * @param availableCourses              The list of courses to iterate through
+   * @param currentSchedule               The current scheduled course list to be added/operated on
+   * @param invariants                    The invariants that constrain the valid schedules
+   * @param generatedScheduledCourseLists The set of scheduledCourse lists that is being generated/added to.
    */
   private void generateSchedulesHelper(int index, List<Course> availableCourses, List<ScheduledCourse> currentSchedule,
       Invariants invariants, Set<List<ScheduledCourse>> generatedScheduledCourseLists) {
@@ -117,7 +118,7 @@ public final class Scheduler {
           if (doesNotOverlapSchedule) {
             List<ScheduledCourse> newScheduledCourseList = new ArrayList<>(currentSchedule);
             newScheduledCourseList.add(new ScheduledCourse(course, lectureSection));
-            generateSchedulesHelper(index +1, availableCourses, 
+            generateSchedulesHelper(index + 1, availableCourses, 
                 newScheduledCourseList, invariants, generatedScheduledCourseLists);
           }
         }
--- a/src/main/java/com/google/sps/Scheduler.java
+++ b/src/main/java/com/google/sps/Scheduler.java
@@ -38,13 +38,13 @@ public final class Scheduler {
     float reqCredits = totalCredits(requiredCourses);
     Set<List<ScheduledCourse>> requiredSet = new HashSet<>();
 
-    generateSchedulesHelper(requiredCourses, new ArrayList<>(), 
+    generateSchedulesHelper(0, requiredCourses, new ArrayList<>(), 
         new Invariants(reqCredits, reqCredits), requiredSet);
     
     // Using those Schedules as a building block, generate all valid schedules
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -297,4 +325,49 @@ public void nonRequiredWithLabBothIncluded() {
 
     Assert.assertEquals(expected, actual);
   }
+
+  public List<Course> createTrash() {
```
achillesecos: Nit: Do you still wanna commit the test to `SchedulerTest.java`? If so, maybe change name lol :)

achillesecos (approved):
LGTM pending comment

achillesecos (commented):
Perhaps it would be a good idea to create an extra test (with a wide invariant range) with 3 courses, that are all non-required, to make sure the `Scheduler` generates correct schedules.

Could be something like:
``` 
     Course 1 = A
     Course 2 = B
     Course 3 = C
```
7 total schedules:
`A, B, C, A & B, A & C, B & C, A & B & C`

src/test/java/com/google/sps/SchedulerTest.java
maceonthompson: haha thanks

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -64,7 +64,7 @@
    * @param invariants  The invariants that constrain the valid schedules
    * @param courseLists The set of course lists that is being generated/added to.
    */
-  private void generateSchedulesHelper(List<Course> availableCourses, List<ScheduledCourse> currentSchedule,
+  private void generateSchedulesHelper(int index, List<Course> availableCourses, List<ScheduledCourse> currentSchedule,
```
kristen-law, line 68: Please update the comment for this function

src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -54,6 +55,33 @@ public void setUp() {
     return Arrays.asList(mon, wed, fri);
   }
 
+  /**
+   * Returns a list of times on monday and friday, to be used in section constructors.
+   */
+  public List<TimeRange> monFri(int hour, int minute, int durationMinutes) {
```
kristen-law, line 61: These day/time helper functions are a bit repetitive.

Instead of defining separate functions for each combination, you could create variables:

```
private static final int[] MON_WED = {TimeRange.MONDAY, TimeRange.WEDNESDAY};
...
```
And then pass those days into a single function:

```
private List<TimeRange> createSectionTimes(int[] days, int hour, int minute, int durationMinutes) {
  List<TimeRange> timeRanges = new ArrayList<>();
  for (int day : days) {
    TimeRange timeRange = TimeRange.fromStartDuration(day, hour, minute, durationMinutes);
    timeRanges.add(timeRange);
  }
  return timeRanges;
}
```

src/test/java/com/google/sps/SchedulerTest.java
```diff
@@ -297,4 +325,49 @@ public void nonRequiredWithLabBothIncluded() {
...
+    "15210", "Computer Science", 12, true, section1, recitation1);
+
+    // RESEARCH AND INNOVATION
+    List<Section> section2 = Arrays.asList(new Section("Professor B", monFri(11, 30, DURATION_90_MINUTES)),
+                                           new Section("Professor B", wedFri(11, 30, DURATION_90_MINUTES)));
+    List<Section> recitation2 = Arrays.asList();
+    Course course2 = new Course("Research and Innovation in Computer Science", "15300", 
+    "Computer Science", 9, true, section2, recitation2);
+
+    // COMPILIERS
```
kristen-law: nit: compiliers -> compilers

src/main/java/com/google/sps/Scheduler.java
```diff
@@ -117,7 +117,7 @@ private void generateSchedulesHelper(List<Course> availableCourses, List<Schedul
           if (doesNotOverlapSchedule) {
             List<ScheduledCourse> newScheduledCourseList = new ArrayList<>(currentSchedule);
             newScheduledCourseList.add(new ScheduledCourse(course, lectureSection));
-            generateSchedulesHelper(new ArrayList<>(availableCourses), 
+            generateSchedulesHelper(index +1, availableCourses,
```
kristen-law: nit: add space after `+`

kristen-law (changes requested):

src/test/java/com/google/sps/SchedulerTest.java
maceonthompson: I agree, this definitely could use some cleaning up. However, given where we're at w/ development & time restrictions, it might be better to focus dev time on finishing features/fixing bugs instead of cleaning up tests (especially since these particular tests will effectively not change from here on out). What are your thoughts?

kristen-law: Sure, just wanted to point out that the five helper functions could be replaced with a single one. I understand the time constraints however so I'll approve this as is.

kristen-law (approved):

PR #40, opened 2020-07-24
grandchild components for the input form

hjohns28:


maceonthompson (approved):
LGTM

frontend/src/Section.js
```diff
@@ -0,0 +1,104 @@
...
+
+  handleDeleteSection() {
+    this.props.deleteSection(this.props.id);
+  }
+
+  render() {
+    return (
+      <Card>
+        <CardContent>
+          <Input placeholder="Professor" inputProps={{ 'aria-label': 'description' }}
```
kristen-law: The `aria-label` here should act as a label for this Input element.

https://www.aditus.io/aria/aria-label/

frontend/src/SubjectPreference.js
```diff
@@ -0,0 +1,27 @@
...
+
+  handleSubjectChange(event) {
+    this.props.updateSubjectPreference(event.target.value)
+  }
+
+  render() {
+    return (
+      <CardContent>
+        <div id="label">Preferred subject: </div>
+        <Input placeholder="Subject" inputProps={{ 'aria-label': 'description' }}
```
kristen-law: Same comment for this `aria-label`

frontend/src/NumberFormat.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import NumberFormat from 'react-number-format';
+import PropTypes from 'prop-types';
+
+export function NumberFormatCustom(props) {
+  const { inputRef, onChange, ...other } = props;
+
+  return (
+    <NumberFormat
```
kristen-law, line 9: I'm not familiar with the react-number-format library. Is this doing something outside of adding a separator at the thousand marks?

kristen-law (approved):
LGTM after comments have been addressed

frontend/src/NumberFormat.js
hjohns28: It makes it so that you cannot enter anything except for a number (eg if you try to type an "a", it won't show up in the input box)

frontend/src/Section.js
hjohns28: Thanks for seeing this, it's actually an unnecessary left-over field from earlier on so I'm gonna get rid of it

frontend/src/NumberFormat.js
kristen-law: Can't you add `type="number"` to your `TextField` to require a number?

hjohns28: Yep, that worked and is definitely simpler. For some reason I thought that attribute couldn't be applied in this case, but it can. I'll change that in the overall input form branch. Thank you!

PR #42, opened 2020-07-24
add overall input form 

hjohns28:
(includes changes from pr #37 - processing / sending input to backend)


maceonthompson (approved):
LGTM

frontend/src/InputForm.js
```diff
@@ -0,0 +1,546 @@
...
+          startTime: "",
+          endTime: "",
+          days: []
+        }]
+      }],
+      criterion: {
+        timePreferences: [{
+          startTime: "",
+          endTime: ""
+        }],
```
achillesecos, line 29: This is the criteria for restricting times/ user no-class time preferences right? Does the `timePreferences` have the list of times that user wants class or does not want class? 

Wasn't the user going to input in the form the times they dont want class?

hjohns28: this is the blocks of time they don't want class

hjohns28: i haven't made that clear to the user, but i was going to update labeling on the form later on

src/test/java/com/google/sps/JSONTest.java
```diff
@@ -0,0 +1,67 @@
...
+      
+      List<Course> courses = request.getCourses();
+      Invariants invariants = request.getCredits();
+      /*  Old logic
+        Type courseListType = new TypeToken<List<Course>>(){}.getType();
+        List<Course> courses = gson.fromJson(wholeJSON.getAsJsonArray("courses"), courseListType);
+    
+        JsonObject basicInfo = wholeJSON.getAsJsonObject("basicInfo");
+        Invariants invariants = gson.fromJson(basicInfo.get("credits"), Invariants.class);
+      */
```
kristen-law: Is there a reason why you're keeping around the old logic in a comment?

frontend/src/InputForm.js
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  updateCourseRank(id, selected) {
+    var previousSelected = this.state.courses[id].rank;
+    if (previousSelected !== "") {
+      var index = this.selectedRanks.indexOf(previousSelected);
+      this.selectedRanks.splice(index, 1);
+    }
+    if (this.selectedRanks.indexOf(selected) === -1) {
+      this.selectedRanks.push(selected);
```
kristen-law: I'd recommend storing selectedRanks as a set instead of an array. `indexOf` is an O(n) operation whereas set look-up is O(1).

frontend/src/InputForm.js
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  updateRankSelectOptions(id) {
+    let items = [];   
+    let numCourses = this.state.courses.length;     
+    if (this.state.courses[id].rank !== "") {
+      for (let j = 1; j <= numCourses; j++) {
+        items.push(<MenuItem key={j} value={j}>{j}</MenuItem>)       
+      }
+      return items;
```
kristen-law: Is this doing what you intend? It looks like you're adding all ranks to the available options when the current course's rank is set. What about ranks that have already been selected?

kristen-law (changes requested):

frontend/src/InputForm.js
hjohns28: It works how I intend, but my intention might not be ideal, this was my thinking: If you haven't selected a rank, you can only choose from the available ranks (that haven't been chosen in other courses). If you have selected a rank though, any option is open in case you want to switch the order, and if you select a rank that's been chosen, the other course with that rank will be cleared. (If it weren't this way, once you pick a rank you wouldn't be able to change it). If there seems like a better option though, I'd love to hear it!

hjohns28: I could make it so that there's a "none" option, or similar, for switching around the ranks if you think that'd be better.

src/test/java/com/google/sps/JSONTest.java
hjohns28: Maceo wrote this file, so I'll check with him, but I should probably delete it. Thanks.

frontend/src/InputForm.js
hjohns28: update: we're actually going to change the ranking logic so that it's a "score" from 1-5 of how much you prefer that course, so courses can have the same score

frontend/src/InputForm.js
hjohns28: Good to know, I'll make that improvement!

hjohns28: (Change doesn't apply with new ranking strategy). Also, I am going to make the type="number" change in the next branch, labSections, so that it's applied everywhere

kristen-law (approved):
LGTM

PR #43, opened 2020-07-27
FullCalendar

achillesecos:
Parent: https://github.com/googleinterns/step78-2020/commit/249e9fecfc00bfe1195605055ebb5d1d1fe38c3b

The changes here will display the hard coded schedules from the prototype on FullCalendar. Will need to change that once input form is merged, so that it fetches the data from the input forms.


src/main/java/com/google/sps/Scheduler.java
```diff
@@ -85,7 +85,7 @@ private void generateSchedulesHelper(List<Course> availableCourses, List<Schedul
 
     List<Section> scheduleSections = new ArrayList<>();
     for (ScheduledCourse c : currentSchedule) {
-      scheduleSections.add(c.getLectureSection());
+      scheduleSections.add(c.getlectureSection());
```
maceonthompson: Don't forget to adjust this to be `getLectureSection` instead of `getlectureSection`

frontend/src/script.js
```diff
@@ -0,0 +1,75 @@
...
+var idArray;
+export async function fetchUserCalendar() {
+  // change back to /fullCalendar
+  const response = await fetch('/fullCalendar');
+  const responseJson = await response.json();
+  return responseJson;
+}
+
+// hide / show the appropriate elements on the page 
+function showButtons() {
```
hjohns28: It doesn't seem like you're using all of these methods from the old google calendar display for full calendar - might want to delete the unnecessary ones if so

src/main/java/com/google/sps/server/FullCalendarDisplay.java
```diff
@@ -0,0 +1,52 @@
...
+  
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      // Get the ordered schedules
+      List<Schedule> schedules = ScheduleCreationHandler.createSchedules();
+
+      Gson gson = new Gson();
+      response.setContentType("application/json");
+      response.getWriter().println(gson.toJson(schedules));
```
maceonthompson, line 46: As a general thing, this logic is handled in the frontend branch already. No need to make changes here, but once we merge the json returne/fetch request will be to `/handleUserInput` (`ScheduleCreationHandler.java`) servlet. Again, no need to change this for now tho.

hjohns28 (approved):
LGTM pending comment (plus there are some console.log() statements you could remove)

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -70,7 +70,7 @@ public void addSchedule(Schedule schedule) throws IOException {
 
       // for each course in the schedule, add all section times to calendar
       for (ScheduledCourse currentCourse : schedule.getCourses()) {
-        List<TimeRange> lectureSectionTimes = currentCourse.getLectureSection().getMeetingTimes();
+        List<TimeRange> lectureSectionTimes = currentCourse.getlectureSection().getMeetingTimes();
```
maceonthompson: See comment in `Scheduler.java`, `getlectureSection` &rightarrow; `getLectureSection` (at least for the final pr into master)

src/test/java/com/google/sps/RankTest.java
```diff
@@ -341,18 +355,88 @@ public void mixedPreferencesDoubleTie() {
...
+    // RESTORE AFTER PRINTING 
+    // Preferences preferenceList = new Preferences(Arrays.asList());
+
+    // List<Schedule> schedules = Arrays.asList(bigSchedule1, bigSchedule2, bigSchedule3, bigSchedule4);
+    // preferenceList.sortSchedules(schedules);
+
+    // List<Schedule> actual = schedules;
+    // List<Schedule> expected = Arrays.asList(bigSchedule1, bigSchedule2, bigSchedule3, bigSchedule4);
+    // Assert.assertEquals(expected, actual);
+
```
maceonthompson, line 647: After merging, delete unused commented out code

maceonthompson (approved):
LGTM pending the getLectureSection change/a few nits!

frontend/src/script.js
achillesecos: This is weird, I had removed these functions... Maybe its on a different commit?

src/main/java/com/google/sps/Scheduler.java
achillesecos: Thanks for the catch. Pretty sure its updated in a different commit, I'll check it out.

src/main/java/com/google/sps/server/FullCalendarDisplay.java
achillesecos: Yea that sounds good. Once we merge things together, we could redirect the fetch to the `/handleUserInput`

src/main/java/com/google/sps/server/ScheduleCalendar.java
achillesecos: Thanks again. I'll find the correct commit that has these changes

src/test/java/com/google/sps/RankTest.java
achillesecos: Sounds good.

frontend/src/script.js
```diff
@@ -15,17 +15,12 @@
...
-      document.getElementById("loading").innerText = "Sorry, looks like there are no possible schedules!"
-    } else {
-      displayCalendar(calIds[0]);
-      showButtons();
-    }
-  });
+
+export async function fetchUserCalendar() {
+  // change back to /fullCalendar
+  const response = await fetch('/fullCalendar');
```
kristen-law: Should this be `'/displayFullCalendar'`?

frontend/src/Calendar.js
```diff
@@ -2,19 +2,109 @@ import React from 'react'
...
+    }
+    //console.log(combinedMeetingTimes)
+    this.setState({
+      schedulesTimes:  [...this.state.schedulesTimes, combinedMeetingTimes]
+    })
+    //return combinedMeetingTimes
+  }
+  
+  getScheduleSections(courseName, meetingTimes) {
+    const lastSunday = this.getLastMonday(Date.now())
```
kristen-law: The variable and function names here are inconsistent - I'm guessing one of them should be changed to Sunday/Monday?

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
 import FullCalendar from '@fullcalendar/react'
 import dayGridPlugin from '@fullcalendar/daygrid'
 
+import moment from 'moment'
+import timeGridPlugin from '@fullcalendar/timegrid'
+
+import {fetchUserCalendar} from './script'
+
+// the conversion of one minute to 60000 milliseconds
+const MIN_TO_MS = 60000;
```
kristen-law, line 11: No need to change anything, but you can avoid the need to add this as a separate constant/comment by documenting the unit in the variable name (e.g. startTimeMs). You'll also see `meetingTime.start * 60 * 1000` sometimes to more explicitly show the calculation.

kristen-law (changes requested):

frontend/src/Calendar.js
achillesecos: Thanks for pointing this out. I changed it to Sunday, since this is technically the start of the week, rather than Monday.

frontend/src/Calendar.js
achillesecos: Thanks for pointing this out.

frontend/src/App.js
```diff
@@ -4,7 +4,9 @@ import Calendar from './Calendar.js';
 
 function App() {
   return (
-    <Calendar />
+    <div>
+      <Calendar />
+    </div>
```
FabriceCastel: Is there a reason for this div? It doesn't seem to do anything here.

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+
 export default class Calendar extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      fetchDone: false,
+      schedulesTimes: [],
+      scheduleIdx: 0     // Current schedule
```
FabriceCastel: nit: `selectedScheduleId` or `selectedScheduleIndex` or similar

-> id or index over idx
-> being more specific about what this schedule index really means

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+
+  componentDidMount(){
+    this.getAllSchedules()
+  }
+
+  async getAllSchedules() {
+    const schedules = await fetchUserCalendar()
+    for(const schedule of schedules) {
+      this.getMeetingTimes(schedule)
+    }
```
FabriceCastel: Change not necessary, but JS gives you a more concise way to express this kind of for-loop:

```
schedules.forEach(this.getMeetingTimes);
```

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+    if(this.state.scheduleIdx < this.state.schedulesTimes.length - 1) {
+      this.setState({scheduleIdx : this.state.scheduleIdx + 1})
+    }
+  }
+
+  getPrevSchedule() {
+    if(this.state.scheduleIdx > 0) {
+      this.setState({scheduleIdx : this.state.scheduleIdx - 1})
+    }
+  }
```
FabriceCastel: These methods aren't getters, they'd be more aptly named `selectNext/PreviousSchedule`

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
-        weekends={false}
-        events={[
-            { title: 'event 1', date: '2019-04-01' },
-            { title: 'event 2', date: '2019-04-02' },
-            { title: getScheduleTitles(), date: getScheduleDates()}
-        ]}
-    />
+      <div>
+        <button onClick = {() => {this.getPrevSchedule()}}>Previous Schedule</button>
+        <button onClick = {() => {this.getNextSchedule()}}>Next Schedule</button>
```
FabriceCastel: If you have a function `foo()`, then

```
() => { foo(); }
() => foo()
foo
```

Are all equivalent. You can replace these with, eg. `onClick={this.getNextSchedule}` and it will avoid instantiating an extra unnecessary lambda.

Additionally, please do not use spaces between properties and their values when declaring HTML elements, ie. `onClick={...}` and not `onClick = {...}`

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+import {fetchUserCalendar} from './script'
+
+// the conversion of one minute to 60000 milliseconds
+const MIN_TO_MS = 60000;
+
 export default class Calendar extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      fetchDone: false,
```
FabriceCastel: Instead of tracking whether the fetch completed in the state, you could check to see if the current schedule index is `< schedulesTimes.length`. It would mean you have less information to keep track of and your render method won't break if you happen to fetch zero calendars from the backend.

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+    }
+  }
+
+  getPrevSchedule() {
+    if(this.state.scheduleIdx > 0) {
+      this.setState({scheduleIdx : this.state.scheduleIdx - 1})
+    }
+  }
+
+  getMeetingTimes(schedule) {
```
FabriceCastel: This function name doesn't reflect what the function does

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+      }
+      combinedMeetingTimes = combinedMeetingTimes.concat(lectureTime)
+    }
+
+    this.setState({
+      schedulesTimes:  [...this.state.schedulesTimes, combinedMeetingTimes]
+    })
+  }
+  
+  getScheduleSections(courseName, meetingTimes) {
```
FabriceCastel: `getScheduleSections` -> `generateCourseMeetingTimes` or something more indicative of what this does

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+    for(const schedule of schedules) {
+      this.getMeetingTimes(schedule)
+    }
+    
+    this.setState({
+      fetchDone: true
+    })
+  }
+
+  getNextSchedule() {
```
FabriceCastel: -> `selectNextSchedule()`

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+    if(this.state.scheduleIdx < this.state.schedulesTimes.length - 1) {
+      this.setState({scheduleIdx : this.state.scheduleIdx + 1})
+    }
+  }
+
+  getPrevSchedule() {
```
FabriceCastel: -> `selectPreviousSchedule()`

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+      // Lab Sections
+      if (course.labSection) {
+        const labTime = this.getScheduleSections(course.name, course.labSection.meetingTimes)
+        combinedMeetingTimes = combinedMeetingTimes.concat(labTime)
+      }
+      combinedMeetingTimes = combinedMeetingTimes.concat(lectureTime)
+    }
+
+    this.setState({
+      schedulesTimes:  [...this.state.schedulesTimes, combinedMeetingTimes]
```
FabriceCastel: `schedulesTimes:  this.state.schedulesTimes.concat(combinedMeetingTimes)`

frontend/src/Calendar.js
```diff
@@ -2,19 +2,108 @@ import React from 'react'
...
+        <button onClick = {() => {this.getPrevSchedule()}}>Previous Schedule</button>
+        <button onClick = {() => {this.getNextSchedule()}}>Next Schedule</button>
+        {this.state.fetchDone &&
+        (<FullCalendar
+          plugins={[ dayGridPlugin, timeGridPlugin ]}
+          initialView="timeGridWeek"
+          events={
+            this.state.schedulesTimes[this.state.scheduleIdx]      
+          }
+        />)}
```
FabriceCastel: nit: indentation

FabriceCastel (changes requested):

frontend/src/script.js
achillesecos: Done.

frontend/src/App.js
achillesecos: True, I used to have buttons in there, but forgot to remove the div when I got rid of them.

frontend/src/Calendar.js
achillesecos: Thanks for the suggestion. I went with `selectedScheduleId`

frontend/src/Calendar.js
achillesecos: Done

frontend/src/Calendar.js
achillesecos: Done

frontend/src/Calendar.js
achillesecos: Changed name to `generateScheduleMeetingTimes() `

frontend/src/Calendar.js
achillesecos: Done. Changed name.

frontend/src/Calendar.js
achillesecos: I believe I tried using the `concat` method, but it did not work.

frontend/src/Calendar.js
achillesecos: Thanks. Changed to those names.

frontend/src/Calendar.js
achillesecos: I see what you mean. Originally I tried that but was running into [this](https://www.andrewconnell.com/blog/deal-with-undefined-this-react-event-handler-performant-way/) issue. I will also fix the `onClick=` spacing error

frontend/src/Calendar.js
achillesecos: Got it. Thank you

frontend/src/Calendar.js
achillesecos: This is a better approach. I changed it to `this.state.selectedScheduleId < this.state.schedulesTimes.length`

frontend/src/Calendar.js
achillesecos: Fixed.

frontend/src/Calendar.js
FabriceCastel: Is `combinedMeetingTimes` an array? If yes, then that would explain it. If not, then `concat()` should return a new array containing everything in the first array, plus the element you gave it as a param

FabriceCastel: The reason this matters is that `concat` behaves differently if the param you give it is an array.

```
const oneTwo = [1, 2];
const threeFour = [3, 4];
const three = 3;

oneTwo.concat(threeFour) => [1, 2, 3, 4]
[ ...oneTwo, threeFour ] => [1, 2, [3, 4]]
[ ...oneTwo, ...threeFour ] = [1, 2, 3, 4]

oneTwo.concat(three) => [1, 2, 3]
[ ...oneTwo, three ] = [1, 2, 3]
```

achillesecos: Yes it is an array, so that explains it

frontend/src/Calendar.js
FabriceCastel: While this solution works, you should consider binding these functions in your component's constructor instead. If you do, then what I've suggested will work just fine and will save you the overhead of instantiating new lambdas.

achillesecos: Got it. I binded the functions into the components constructor [here](https://github.com/googleinterns/step78-2020/pull/43/commits/d82d8120e0cfa8debdafaee44bbc573ea79d61ca) and this should save the overhead you mentioned.

FabriceCastel (approved):
LGTM - please run the linter again to fix any pending style issues from your last commits before merging.

kristen-law (approved):

PR #44, opened 2020-07-27
update input form to have recitation sections

hjohns28:
(I also updated the value of some grandchild components so they would show the proper value when a component - section, course, or time preference - was deleted)


maceonthompson (approved):
Like we talked about, avoiding the duplicate code in LabSection/LectureSection.js would be ideal, but for now this LGTM! 
Hopefully Kristen or Fabrice will have an idea as to how to avoid it, but worst case we can implement that fix down the line

frontend/src/Course.js
```diff
@@ -123,18 +153,52 @@ class Course extends React.Component {
...
+              professor={this.props.lectureSections[index].professor}
+              startTime={this.props.lectureSections[index].startTime}
+              endTime={this.props.lectureSections[index].endTime}
+              sun={this.props.lectureSections[index].days.indexOf(0) !== -1}
+              mon={this.props.lectureSections[index].days.indexOf(1) !== -1}
+              tue={this.props.lectureSections[index].days.indexOf(2) !== -1}
+              wed={this.props.lectureSections[index].days.indexOf(3) !== -1}
+              thur={this.props.lectureSections[index].days.indexOf(4) !== -1}
+              fri={this.props.lectureSections[index].days.indexOf(5) !== -1}
+              sat={this.props.lectureSections[index].days.indexOf(6) !== -1}
```
kristen-law: Consider storing `days` in a set instead of an array since `indexOf` is an O(n) operation

frontend/src/LabSection.js
```diff
@@ -8,7 +8,7 @@ import Checkbox from '@material-ui/core/Checkbox';
 import IconButton from '@material-ui/core/IconButton';
 import DeleteIcon from '@material-ui/icons/Delete';
 
-class Section extends React.Component {
+class LabSection extends React.Component {
```
kristen-law, line 11: Agreed that we should avoid creating a duplicate class for lab sections. Unfortunately, I don't know enough about React to recommend the best way to avoid that - maybe Fabrice can suggest a solution when he's back tomorrow.

frontend/src/InputForm.js
```diff
@@ -439,11 +559,14 @@ class InputForm extends React.Component {
...
     let submitState = JSON.parse(JSON.stringify(this.state));
-    submitState.courses = submitState.courses.map((course) => this.convertCourseSections(course));
+    submitState.courses = submitState.courses.map((course) => this.convertCourseLectureSections(course));
+    submitState.courses = submitState.courses.map((course) => this.convertCourseLabSections(course))
 
     submitState.criterion.timePreferences = submitState.criterion.timePreferences.map((times) => 
         this.timeToTimeRange(0, times.startTime, times.endTime));
 
+    console.log(JSON.stringify(submitState))
+
```
kristen-law: Remove?

frontend/src/InputForm.js
```diff
@@ -155,76 +164,150 @@ class InputForm extends React.Component {
...
-              ? ({...section, endTime: sectionEndTime})
-              : section)})
+              ? ({...labSection, endTime: sectionEndTime})
+              : labSection)})
           : course)
     }))
   }
 
-  updateSectionDays(courseID, sectionID, day) {
+  updateLabSectionDays(courseID, sectionID, day) {
```
kristen-law, line 330: I just noticed the nested ternary operators in this file. I'd really recommend changing these to if-else statements - they are very difficult to follow as nested ternary operators.

kristen-law (changes requested):

frontend/src/Course.js
hjohns28: Right, I will make that change, thanks!

frontend/src/InputForm.js
hjohns28: Yeah I see that it did get more unreadable when nested, I'll clear that up!

frontend/src/InputForm.js
```diff
@@ -0,0 +1,791 @@
...
+    this.setState({
+      ...this.state,
+      courses: this.state.courses.map((course, index) =>
+        index === id
+          ? ({ ...course, isRequired: !this.state.courses[index].isRequired })
+          : course)
+    });
+  }
+
+  updateCourseRank(id, selected) {
```
hjohns28: (Didn't change nested ternary operators in udpateCourseRank() because they'll go away with the new scoring changes in the other branch, grandparentComponents)

kristen-law (approved):

PR #50, opened 2020-07-29
Add stepper with buttons to click between steps

hjohns28:


frontend/src/Schedulr.js
```diff
@@ -0,0 +1,83 @@
...
+  getStepContent(step) {
+    switch (step) {
+    case 0:
+      return (<InputForm
+        handleNext={this.handleNext}
+        setScheduleList={this.setScheduleList}/>);
+    case 1:
+      return (<p>{JSON.stringify(this.state.scheduleList)}</p>);
+    case 2:
+      return 'This is the bit I really care about!';
```
maceonthompson, line 33: Note: the logic for step 3 (case 2) will be the exporting to google calendar - this will be implemented at a later date.

maceonthompson (approved):
lgtm! (Though @ whichever host reviews this, this might be a little cheat-y as I contributed to some of this branch)

achillesecos (approved):
LGTM

frontend/src/Schedulr.js
```diff
@@ -0,0 +1,83 @@
...
+    return (
+      <div>
+        <Stepper activeStep={this.state.activeStep}>
+          {this.state.steps.map((label, index) => {
+            const stepProps = {};
+            const labelProps = {};
+            return (
+              <Step key={label} {...stepProps}>
+                <StepLabel {...labelProps}>{label}</StepLabel>
+              </Step>
```
kristen-law: `stepProps` and `labelProps` can be removed here if you're not adding anything to them.

frontend/src/InputForm.js
```diff
@@ -630,15 +630,15 @@ class InputForm extends React.Component {
...
     fetch('/handleUserInput', {
       method: 'POST',
       headers: {'Content-Type': 'application/json'},
       body: JSON.stringify(submitState),
     }).then((response) => response.json())
       .then((responseSchedules) => {
-        scheduleList = responseSchedules;
-        console.log(JSON.stringify(scheduleList));
+        this.props.setScheduleList(responseSchedules);
+        console.log(JSON.stringify(responseSchedules));
```
kristen-law: Remove logging

kristen-law (approved):
LGTM, just some small comments

frontend/package.json
```diff
@@ -36,7 +38,6 @@
     ]
   },
   "devDependencies": {
-    "eslint": "^7.5.0",
```
FabriceCastel, line 39: This deletion seems like a mistake?

frontend/src/Schedulr.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  getStepContent(step) {
+    switch (step) {
+    case 0:
+      return (<InputForm
+        handleNext={this.handleNext}
+        setScheduleList={this.setScheduleList}/>);
+    case 1:
+      return (<p>{JSON.stringify(this.state.scheduleList)}</p>);
```
FabriceCastel: Tip: try out `JSON.stringify(this.state.scheduleList, null, 2)` instead. It'll format the JSON string nicely.

frontend/package.json
maceonthompson: This was intentional: Create-React App already includes an Eslint dependency - having this included in package.json causes errors when running `npm run build` and `npm run start`. Even with the deletion `npm install` should still install eslint? Although we aren't 100% sure.

frontend/src/Schedulr.js
```diff
@@ -0,0 +1,83 @@
...
+            const labelProps = {};
+            return (
+              <Step key={label} {...stepProps}>
+                <StepLabel {...labelProps}>{label}</StepLabel>
+              </Step>
+            );
+          })}
+        </Stepper>
+        <div>
+          <Typography>{this.getStepContent(this.state.activeStep)}</Typography>
```
FabriceCastel: If you only ever call `getStepContent()` with the same param, then you can instead make change it so it doesn't require any params:

```
getStepContent() {
  const step = this.state.activeStep;
  ...
}
```

And here, instead `<Typography>{this.getStepContent()}</Typography>`

frontend/src/Schedulr.js
hjohns28: Good to know, thank you !

frontend/package.json
FabriceCastel: Would be good to sanity check this with `rm -rf node_modules && npm install` and then running the linter to see if it still works as expected.

frontend/src/Schedulr.js
hjohns28: That makes sense, thanks for seeing that.

frontend/package.json
maceonthompson: Good call - it still works!

PR #51, opened 2020-07-30
Tests fix

achillesecos:
Parent: https://github.com/googleinterns/step78-2020/commit/515584d3a569f96b6b576f8c5441cf63ecfbc200

Deleted some old functions of `RankTest` that were added to master


hjohns28 (approved):
LGTM

src/main/java/com/google/sps/server/FullCalendarDisplay.java
```diff
@@ -1,51 +0,0 @@
-/*
```
kristen-law, line 1: Did you mean to delete this file?

achillesecos: Yes. This servlet was originally used to give FullCalendar the hardcoded schedules, but now since we have the Stepper in the front end, we're working on passing the input form data instead.

kristen-law (approved):
LGTM

PR #52, opened 2020-07-30
Integrate FullCalendar into stepper

hjohns28:


maceonthompson (approved):
LGTM!

frontend/src/Schedulr.js
```diff
@@ -28,15 +35,20 @@ class Schedulr extends React.Component {
         handleNext={this.handleNext}
         setScheduleList={this.setScheduleList}/>);
     case 1:
-      return (<p>{JSON.stringify(this.state.scheduleList, null, 2)}</p>);
+      return (<Calendar 
+        scheduleList={this.state.scheduleList}
+        schedulesTimes={this.state.schedulesTimes}/>);
```
achillesecos: Might wanna run lint fix on this

frontend/src/Schedulr.js
```diff
@@ -48,6 +60,58 @@ class Schedulr extends React.Component {
     this.setState({...this.state, scheduleList: listOfSchedules});
   }
 
+  generateSchedulesTimes() {
+    //TODO: This is currently broken. Figure out why.
+    //console.log('generating schedules: ${JSON.stringify(this.props.scheduleList)}');
```
achillesecos: Still need this?

achillesecos (approved):
lgtm pending comments

frontend/src/Schedulr.js
```diff
@@ -48,6 +60,56 @@ class Schedulr extends React.Component {
     this.setState({...this.state, scheduleList: listOfSchedules});
   }
 
+  generateSchedulesTimes() {
+    this.state.scheduleList.forEach((schedule) => {
+      this.generateScheduleMeetingTimes(schedule);
+    });
+    return true;
```
FabriceCastel, line 67: You should consider adding an eslint rule for line breaks around code blocks

https://eslint.org/docs/rules/padding-line-between-statements

```
"padding-line-between-statements": [
  { "blankLine": "always", "prev": "block-like", "next": "*" },
  { "blankLine": "always", "prev": "*", "next": "block-like" }
]
```

I think that rule might work. Haven't tested it though, so YYMV but the point is to enforce new lines before and after code blocks, eg.

```
function foo() {
  const a = 2;
  if (a > 3) {
    console.log('oh no');
  }
  return 'apple';
}
```

Would become

```
function foo() {
  const a = 2;

  if (a > 3) {
    console.log('oh no');
  }

  return 'apple';
}
```

This could be done in a short follow-up PR

FabriceCastel (approved):
LGTM - please take note of the comment and follow up on it in a separate PR

frontend/src/Schedulr.js
hjohns28: Sounds good, we will add that, thank you!

PR #55, opened 2020-07-31
integrate google calendar

hjohns28:


src/main/java/com/google/sps/server/ScheduleDisplayHandler.java
```diff
@@ -26,36 +26,42 @@
...
 import com.google.api.services.calendar.Calendar;
 import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import com.google.sps.data.*;
 
 /**
  * Exports the ordered schedules to secondary calendars on the user's Google calendar    
  */
 @WebServlet("/handleSchedules")
```
maceonthompson: Since this is used for exporting to gcal, it might be worth renaming to something more descriptive

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -115,46 +107,33 @@ public String getCalendarId() {
...
    *
    * @param course the course being added 
-   * @param startTime the start time for the calendar event
-   * @param endTime the end time for the calendar event
+   * @param startTime the start time of the course
+   * @param endTime the end time of the course
+   * @param termLength the number of weeks in the user's college term
    */
-  private void addEvent(ScheduledCourse course, String startTime, String endTime) throws IOException {
+  private void addEvent(Course course, ZonedDateTime startTime, ZonedDateTime endTime, int termLength) throws IOException {
```
maceonthompson: If Course is only being used for it's name, it might be worth just passing in the name of the course as a string (feel free to not make this change if you plan on using more course info down the line)

maceonthompson (changes requested):

src/main/java/com/google/sps/server/ScheduleCalendar.java
```diff
@@ -62,46 +68,32 @@ public ScheduleCalendar(com.google.api.services.calendar.Calendar client) throws
...
-      // for each course in the schedule, add all section times to calendar
+      // for each course in the schedule, add an event to the calendar
       for (ScheduledCourse currentCourse : schedule.getCourses()) {
-        List<TimeRange> lectureSectionTimes = currentCourse.getLectureSection().getMeetingTimes();
-        
-        addSectionToCalendar(lectureSectionTimes, calendar, currentCourse);
-        if (currentCourse.getLabSection() != null) {
-          List<TimeRange> labSectionTimes = currentCourse.getLabSection().getMeetingTimes();
-          addSectionToCalendar(labSectionTimes, calendar, currentCourse);
+        List<TimeRange> sectionTimes = currentCourse.getLectureSection().getMeetingTimes();
```
achillesecos: Do you also add the `labSections` of the given course?

src/main/java/com/google/sps/server/ScheduleDisplayHandler.java
hjohns28: That's a good point, I'll update the name!

src/main/java/com/google/sps/server/ScheduleCalendar.java
hjohns28: I was thinking of expanding the fields for calendar events, so I'm not going to change it right now, but good catch. Thanks!

achillesecos (approved):
LGTM pending comments

src/main/java/com/google/sps/server/ScheduleCalendar.java
hjohns28: I forgot to add the labSections in - thanks for seeing that!

maceonthompson (approved):
LGTM

achillesecos (approved):
LGTM

frontend/src/InputForm.js
```diff
@@ -634,6 +634,7 @@ class InputForm extends React.Component {
     }).then((response) => response.json())
       .then((responseSchedules) => {
         this.props.setScheduleList(responseSchedules);
+        this.props.setTermDates(this.state.basicInfo.termDates.startDate, this.state.basicInfo.termDates.endDate);
```
FabriceCastel: nit: improve legibility here by destructuring the state first:

```
const { startDate, endDate } = this.state.basicInfo.termDates;

...

this.props.setTerDates(startDate, endDate);
```

or something similar

frontend/src/Schedulr.js
```diff
@@ -15,38 +15,55 @@ class Schedulr extends React.Component {
...
+    this.exportToGoogleCalendar = this.exportToGoogleCalendar.bind(this);
+    this.setTermDates = this.setTermDates.bind(this);
   }
 
   getStepContent() {
     const step = this.state.activeStep;
     switch (step) {
     case 0:
+      return (
+        <Button onClick={this.signIn}>Get Started</Button>
```
FabriceCastel: Would it make more sense for this button to say something to the effect of "sign in" with some additional text giving the context around this (ie. we need you to give us google calendar access in order to export the calendars).

frontend/src/Schedulr.js
```diff
@@ -15,38 +15,55 @@ class Schedulr extends React.Component {
...
+    case 2:
       return (<Calendar
         scheduleList={this.state.scheduleList}
-        schedulesTimes={this.state.schedulesTimes}/>);
-    case 2:
-      return 'Export to google calendar: coming soon ;)';
+        schedulesTimes={this.state.schedulesTimes}
+        exportToGoogleCalendar={this.exportToGoogleCalendar}/>);
+    case 3:
+      return (<iframe title="google calendar" id="calendar" src={this.state.calId} width="800" height="600" frameBorder="0" scrolling="no"></iframe>);
```
FabriceCastel:
```
return (
  <iframe
    title="google calendar"
    id="calendar"
    src={this.state.calId}
    width="800"
    height="600"
    frameBorder="0"
    scrolling="no">
  </iframe>);
```

or similar with line breaks for legibility

frontend/src/Schedulr.js
```diff
@@ -15,38 +15,55 @@ class Schedulr extends React.Component {
...
     default:
       return 'Unknown step';
     }
   }
 
+  signIn() {
+    window.location.assign('https://course-scheduler-step-2020.uc.r.appspot.com/auth');
```
FabriceCastel: You shouldn't hard-code the URL here. If you deploy this site at a different address, it will break. What you really want is a relative path:

```
window.location.assign('/auth');
```

src/main/java/com/google/sps/server/Auth.java
```diff
@@ -33,7 +33,7 @@
   @Override
   protected void doGet(HttpServletRequest request, HttpServletResponse response)
       throws IOException {
-    response.sendRedirect("/course-form.html");
+    response.sendRedirect("/?signedIn=true");
```
FabriceCastel, line 36: While this URL param trick will mostly work ok, it's not reliable. Anyone can go and type that param in the URL bar and the frontend will behave as though the user had signed in when they hadn't.

You can address this in a follow-up PR if it's too big a change for now, here is a suggestion as to the officially supported way to do this: https://stackoverflow.com/a/50256404

src/main/java/com/google/sps/server/CalendarUtils.java
```diff
@@ -14,105 +14,85 @@
...
+   * @return a LocalTime representing the time for creating the calendar event for the current course
    */
-  public static int getCurrentMonth(Calendar calendar) {
-    return calendar.get(Calendar.MONTH) + 1;
-  }
-  
+  private static LocalTime getCourseTime(int timeInMin, int courseDayOfWeek) {
+    int dayInHours = courseDayOfWeek * 24;
+    int hour = (int) Math.floor((timeInMin / 60) - dayInHours);
+    int minute = (int) (60 * (((timeInMin / 60.0) - dayInHours) - hour));
```
FabriceCastel: Would this achieve the same thing?

```
int minute = timeInMin % 60;
```

src/main/java/com/google/sps/server/CalendarUtils.java
```diff
@@ -14,105 +14,85 @@
...
+   * @param courseDayOfWeek the day of week the current course section starts on 
+   * @return a LocalTime representing the time for creating the calendar event for the current course
    */
-  public static int getCurrentMonth(Calendar calendar) {
-    return calendar.get(Calendar.MONTH) + 1;
-  }
-  
+  private static LocalTime getCourseTime(int timeInMin, int courseDayOfWeek) {
+    int dayInHours = courseDayOfWeek * 24;
+    int hour = (int) Math.floor((timeInMin / 60) - dayInHours);
```
FabriceCastel: Would this achieve the same thing?

```
int hour = (timeInMin / 60) % 24;
```

src/main/java/com/google/sps/server/CalendarUtils.java
```diff
@@ -14,105 +14,85 @@
...
-    int day = dayOfWeek + nextSun;
-     
-    String dayString = day > 9 ? Integer.toString(day) : "0" + Integer.toString(day);
-    String monthString = month > 9 ? Integer.toString(month) : "0" + Integer.toString(month);
-    String hourString = hour > 9 ? Integer.toString(hour) : "0" + Integer.toString(hour);     
-    String minuteString = minute > 9 ? Integer.toString(minute) : "0" + Integer.toString(minute);
-
-    String dateTime = year + "-" + monthString + "-" + dayString + "T" + hourString + ":" + minuteString + ":00";
+  public static ZonedDateTime calculateDateTime(Calendar client, LocalDate startDate, int timeInMin) { 
+    int courseDayOfWeek = (int) Math.floorDiv(timeInMin, 24*60);
```
FabriceCastel: Is this any different than:

```
int courseDayOfWeek = timeInMin / (24 * 60);
```

FabriceCastel (changes requested):

frontend/src/Schedulr.js
hjohns28: Yes, I should definitely give more context - I was going to add more of that in later, but I will do that here. Thanks!

frontend/src/Schedulr.js
hjohns28: Got it, thanks for letting me know!

src/main/java/com/google/sps/server/CalendarUtils.java
hjohns28: yes - that should be the same, not sure how I ended up with such a complicated equation but I definitely should have checked to simplify it. Sorry about that and thanks for noticing!

src/main/java/com/google/sps/server/CalendarUtils.java
hjohns28: Correct again, sorry!

src/main/java/com/google/sps/server/CalendarUtils.java
hjohns28: Nope, again it's the same. Will change!

src/main/java/com/google/sps/server/Auth.java
hjohns28: I evidently should have done more searching about checking the sign in state, but thank you for bringing this to my attention, I really appreciate it and will make sure to be more secure next time. I'm going to open a new PR for this change.

src/main/java/com/google/sps/server/CalendarUtils.java
```diff
@@ -54,8 +54,8 @@ private static LocalDate getCourseDate(LocalDate startDate, int timeInMin, int c
    */
   private static LocalTime getCourseTime(int timeInMin, int courseDayOfWeek) {
     int dayInHours = courseDayOfWeek * 24;
-    int hour = (int) Math.floor((timeInMin / 60) - dayInHours);
-    int minute = (int) (60 * (((timeInMin / 60.0) - dayInHours) - hour));
+    int hour = (int) ((timeInMin / 60) % 24);
+    int minute = (int) (timeInMin % 60);
```
FabriceCastel: I don't think you need the type casting here, these expressions already evaluate to `int` types.

src/main/java/com/google/sps/server/CalendarUtils.java
```diff
@@ -88,7 +88,7 @@ private static ZoneId getZoneId(Calendar client) {
    * @return a ZonedDateTime, to be used in creating a calendar event 
    */
   public static ZonedDateTime calculateDateTime(Calendar client, LocalDate startDate, int timeInMin) { 
-    int courseDayOfWeek = (int) Math.floorDiv(timeInMin, 24*60);
+    int courseDayOfWeek = (int) (timeInMin / (24 * 60));
```
FabriceCastel: Same comment here about type casting

FabriceCastel (approved):
LGTM pending the last two minor comments

src/main/java/com/google/sps/server/CalendarUtils.java
hjohns28: Gotcha that makes sense, whoops, thank you!

PR #56, opened 2020-07-31
Update InputForm to cache state locally

maceonthompson:


hjohns28 (approved):
LGTM

frontend/src/InputForm.js
```diff
@@ -7,37 +7,39 @@ import BasicInfo from './BasicInfo';
...
+          timePreferences: [],
+          preferredSubject: '',
         },
-        termDates: {
-          startDate: '',
-          endDate: '',
+        basicInfo: {
+          credits: {
+            minCredits: 12,
+            maxCredits: 18,
```
achillesecos, line 35: Why are the `minCredits/maxCredits` set? Do they get overwritten?

achillesecos (approved):
LGTM pending comment

frontend/src/InputForm.js
maceonthompson: These have been set as defaults - this isn't a change made in this pull request. If we do want to change it (which makes sense given that the user's min/max credits could definitely be different from the default) it would be worth opening a separate PR.

achillesecos: Ah ok, I forgot it was defaulted before. Up to you, its ok as is.

kristen-law (approved):
LGTM

PR #57, opened 2020-08-01
improve security by checking if user is logged in with auth2

hjohns28:


frontend/src/Schedulr.js
```diff
@@ -175,6 +176,20 @@ class Schedulr extends React.Component {
     }));
   }
 
+  componentDidMount() {   
+    window.gapi.load('auth2', () => {
+      this.auth2 = gapi.auth2.init({
+        client_id: '119851197452-crnk45b5i6gsi5povitstfpd203n7j6b.apps.googleusercontent.com',
```
maceonthompson, line 187: Is it okay for us to include this or do we need to obfuscate it?

maceonthompson (approved):
LGTM pending the comment on Schedulr.js

achillesecos (approved):
LGTM

frontend/src/Schedulr.js
hjohns28: I looked it up and I believe that the client id doesn't need to be hidden (the client secret does obviously), so we should be good, but if needed I will change and obfuscate it. Thanks!

FabriceCastel (approved):
LGTM

PR #62, opened 2020-08-05
resize calendars, change navigation buttons

hjohns28:
Ideally I would like to use a percentage (not fixed size) for the google calendar, and center it, but I haven't figured that out. These are just a couple changes that make the UI a bit prettier.


maceonthompson (approved):
LGTM

kristen-law (approved):
LGTM

